PR #29, opened 2019-07-10
support taking cluster

njzjz:
- Add `use_cluster` and `cluster_cutoff` parameters
  - Detect atoms within the model deviation range and take clusters with a radius of `cluster_cutoff` (temporarily a bit slow)
  - Use `MultiSystems` to convert First Principles results to DeePMD systems and continue next iteration
- Add `init_multi_systems` parameter: allow the init data system to be a folder containing several systems for training
- Add `sys_format` parameter: assign the format of `sys_configs` (default is still POSCAR)
- Fix `LSFJob` bug


dpgen/generator/lib/gaussian.py
```diff
@@ -4,16 +4,32 @@
...
+import dpdata
 from scipy.sparse import csr_matrix
 from scipy.sparse.csgraph import connected_components
+from scipy.spatial import cKDTree
 try:
     import openbabel
 except ImportError:
     pass
+try:
+    from ase import Atoms
```
njzjz: No idea about how to handle coordinates with periodic boundary conditions... So I use a third party module instead.

PR #51, opened 2019-07-25
stop main process by ratio of successing jobs

haidi-ustc:


dpgen/generator/run.py
```diff
@@ -1346,12 +1350,22 @@ def post_fp_vasp (iter_index,
...
         sys_data_path = os.path.join(work_path, 'data.%s'%ss)
         all_sys.to_deepmd_raw(sys_data_path)
         all_sys.to_deepmd_npy(sys_data_path, set_size = len(sys_outcars))
 
+    dlog.info("failed frame number: %s "%icount)
+    dlog.info("total frame number: %s "%tcount)
+    reff=icount/tcount
+    dlog.info('ratio of failed frame:  {:.2%}'.format(reff*100))
+
+    if reff>0.05:
```
amcadmus: user should be able to set this level (by default it is 0.05)

dpgen/generator/run.py
```diff
@@ -1310,11 +1310,15 @@ def post_fp_vasp (iter_index,
...
+    tcount=0 
+    icount=0 
     for ss in system_index :
         sys_outcars = glob.glob(os.path.join(work_path, "task.%s.*/OUTCAR"%ss))
         sys_outcars.sort()                
 
         flag=True
+        tcount+=len(sys_outcars)
         for oo in sys_outcars :
             if flag:
```
amcadmus: check flat within condition len(_sys)>1 and len(_sys) == 1 ?

haidi-ustc: len(_sys) should be 0 or 1  when  all data collected from static calculation step.

PR #85, opened 2019-08-21
fix bug for cp2k

robinzyb:
This version can run the full CH4 example using cp2k_fp


robinzyb:
This version can run the full CH4 example using cp2k_fp

PR #86, opened 2019-08-22
Devel: Enabled the supportation for LSF job management system

Cloudac7:
Enabled the support for LSF system, which is tested preliminarily and worked well. For prevention of the conflict while submitting the jobs, a function for checking the state of queues is added for LSF only and would be easy to migrate to other job management systems.


dpgen/remote/RemoteJob.py
```diff
@@ -731,22 +731,25 @@ def submit(self,
             status = self.check_status()
             if status in [  JobStatus.unsubmitted, JobStatus.unknown, JobStatus.terminated ]:
                 dlog.debug('task restart point !!!')
+                while self.check_limit(task_max=resources['task_max']):
```
njzjz: Can you add `if 'task_max' in resources and resources['task_max']>0:` in front of this line to make the default parameter is unlimited?

dpgen/remote/RemoteJob.py
```diff
@@ -731,22 +731,25 @@ def submit(self,
...
-           #except:
+            #except:
                #dlog.debug('no job_id file')
                #dlog.debug('task restart point !!!')
                #self._submit(job_dirs, cmd, args, resources)
         else:
-           dlog.debug('new task!!!')
-           self._submit(job_dirs, cmd, args, resources)
+            dlog.debug('new task!!!')
+            while self.check_limit(task_max=resources['task_max']):
```
njzjz: So is it.

dpgen/remote/RemoteJob.py
```diff
@@ -821,12 +839,17 @@ def _make_script(self,
...
-        else :
-            ret += '#BSUB -R "select[ngpus >0] rusage[ngpus_excl_p=1]"\n#BSUB -n %d\n' % (res['numb_gpu'])
-        #ret += '#BSUB -l walltime=%s\n' % (res['time_limit'])
-        #if res['mem_limit'] > 0 :
-        #    ret += "#BSUB -l mem=%dG \n" % res['mem_limit']
+            ret += '#BSUB -R span[ptile=%d]\n#BSUB -n %d\n' % (
+                res['node_cpu'], res['numb_node'] * res['task_per_node'])
+        else:
+            ret += '#BSUB -R span[ptile=%d]\n#BSUB -n %d\n' % (
+                res['numb_node'] * res['task_per_node'], res['numb_node'] * res['task_per_node'])
```
njzjz: What the purpose of this line? It originally aims to submit a GPU task. Why not use `numb_gpu`?

njzjz: And, are you sure this script can apply for a GPU resource?

Cloudac7: For the test cluster I finished the test, it is of no use to add `#BSUB -R "select[ngpus >0] rusage[ngpus_excl_p=1]"` item, for the GPU hosts are configured in the same way as CPU ones. And for the original code, it would be useful if the GPU hosts were configured in the right way. So I will modify my code to support the correct situation. Thanks for your comment.

dpgen/remote/RemoteJob.py
Cloudac7: OK, thanks.

PR #108, opened 2019-09-25
use dpdata for pwscf

jameswind:


dpgen/generator/run.py
```diff
@@ -1305,16 +1305,14 @@ def post_fp_pwscf (iter_index,
         flag=True
         for ii,oo in zip(sys_input,sys_output) :
             if flag:
-                _sys = dpdata.LabeledSystem(type_map = jdata['type_map'])
-                _sys.data=cvt_1frame(ii,oo)
+                _sys = dpdata.LabeledSystem(oo, fmt = 'pwscf/scf', type_map = jdata['type_map'])
```
njzjz:
```suggestion
                _sys = dpdata.LabeledSystem(oo, fmt = 'qe/pw/scf', type_map = jdata['type_map'])
```
In addition, it's necessary to update `dpdata` version in `setup.py`.

amcadmus: We should require dpdata  >= 0.1.7 in the setup.py to impose the updated format 'qe/pw/scf'

PR #122, opened 2019-10-05
add an option to skip write out files (fix #109)

### LINKED ISSUE #109: task 06 is slow and needs lots of memory if trajectory is large
Task 06 loads all model deviations in the MD simulation into the memory, and then turns to lists of index. If the data is big, it will occupy lots of memory and consume lots of time to write out files.

TODO: add an option to minimize the memory. (e.g., write out files can be skipped, and it is only necessary to create candidate list)

njzjz:
otherwise it is too slow

### DIFF
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -700,11 +700,14 @@ def _make_fp_vasp_inner (modd_path,
     system_index.sort()
 
     fp_tasks = []
-    cluster_cutoff = jdata['cluster_cutoff'] if 'use_clusters' in jdata and jdata['use_clusters'] else None
+    cluster_cutoff = jdata['cluster_cutoff'] if jdata.get('use_clusters', False) else None
+    # skip save *.out if make_fp_out is False, default is True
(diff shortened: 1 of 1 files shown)


dpgen/generator/run.py
```diff
@@ -700,11 +700,14 @@ def _make_fp_vasp_inner (modd_path,
     system_index.sort()
 
     fp_tasks = []
-    cluster_cutoff = jdata['cluster_cutoff'] if 'use_clusters' in jdata and jdata['use_clusters'] else None
+    cluster_cutoff = jdata['cluster_cutoff'] if jdata.get('use_clusters', False) else None
+    # skip save *.out if make_fp_out is False, default is True
+    make_fp_out = jdata.get("make_fp_out", True)
```
amcadmus: The meaning of "make_fp_out" is not clear.... How about "detailed_report_make_fp" ?  
If the user do not want a detailed report as what we are providing now, a brief report on the percentage of candidate, accurate and failed should be given.

PR #178, opened 2019-11-27
fix the bug when not using the vasp_params

wanghaopku:


amcadmus:
please let us know if pr 180 fixes you issue

wanghaopku:
> please let us know if pr 180 fixes you issue

yes, i have tested it. Thanks.

PR #196, opened 2019-12-21
support nopbc

njzjz:
waiting for deepmd?


amcadmus:
For "use_cluseter", do we have to manually touch `nopbc` in each data system?

njzjz:
> For "use_cluseter", do we have to manually touch `nopbc` in each data system?

yes, but not manually. It can be implemented in https://github.com/deepmodeling/dpdata/pull/70.

PR #201, opened 2020-01-14
add the interface between dpgen and pwmat

HongzhenTian:
add the interface between dpgen and pwmat


amcadmus:
1, Please add unittests for your implementation
2, The PP files for unittests should be empty.
3, Please do not add PP, data files to example. In all examples,  only provide the input json file.

PR #202, opened 2020-01-14
Fix some bugs for LSF system and add key 'sleep' to machine resource

Cloudac7:
There are some known issue using dpgen to work on lsf host, and in this PR, they are fixed. I add the state 
`PSUSP` and `USUSP` to support manually hanging up tasks, instead of resubmitting. And for LSF >= 10.1 with SP6, the gpu tasks could be submitted in a different way, so the key `new_lsf_gpu` was added to the `resources` part of `machine.json`. Also, I found the `sleep` method in `Batch.py`, so I add `sleep` key in `resources` to enable the sleep step after submitting, in prevention of the crash during submitting.


dpgen/dispatcher/LSF.py
```diff
@@ -1,31 +1,36 @@
...
-            if self.check_finish_tag() :
+        if ("Job <%s> is not found" % job_id) in err_str:
+            if self.check_finish_tag():
                 return JobStatus.finished
-            else :
+            else:
                 return JobStatus.terminated
-        elif ret != 0 :
-            raise RuntimeError ("status command bjobs fails to execute. erro info: %s return code %d"
-                                    % (err_str, ret))
```
amcadmus: you should not change the format of the code (even though you do no like it). 

Such changes make people difficult to locate the real change to the code

Cloudac7: Thanks for your advise. I just committed to restore the format of the code to show where I modified. Sorry for using autopep8 directly to format the code without asking in ahead.

dpgen/dispatcher/LSF.py
```diff
@@ -102,14 +102,19 @@ def sub_script_head(self, res):
         ret = ''
         ret += "#!/bin/bash -l\n#BSUB -e %J.err\n#BSUB -o %J.out\n"
         if res['numb_gpu'] == 0:
-            ret += '#BSUB -R span[ptile=%d]\n#BSUB -n %d\n' % (
-                res['node_cpu'], res['numb_node'] * res['task_per_node'])
+            ret += '#BSUB -n %d\n#BSUB -R span[ptile=%d]\n' % (
```
njzjz: Why modify this line? It seems that it’s as same as before.

dpgen/dispatcher/LSF.py
```diff
@@ -102,14 +102,19 @@ def sub_script_head(self, res):
...
+                res['numb_node'] * res['task_per_node'], res['node_cpu'])
         else:
             if res['node_cpu']:
                 ret += '#BSUB -R span[ptile=%d]\n' % res['node_cpu']
-                # It is selected only for the situation that GPU is related to CPU node.
-            ret += '#BSUB -R "select[ngpus >0] rusage[ngpus_excl_p=1]"\n#BSUB -n %d\n' % (
-                res['numb_gpu'])
+            if 'new_lsf_gpu' in res and res['new_lsf_gpu'] == True:
```
njzjz:
```suggestion
            if res.get('new_lsf_gpu', False):
```

dpgen/remote/RemoteJob.py
```diff
@@ -770,6 +769,7 @@ def _submit(self,
         with sftp.open(os.path.join(self.remote_root, 'job_id'), 'w') as fp:
             fp.write(job_id)
         sftp.close()
+        time.sleep(20) # For preventing the crash of the tasks while submitting.
```
njzjz: I am not sure but it seems RemoteJob.py has been deprecated? @amcadmus

dpgen/dispatcher/Batch.py
```diff
@@ -119,7 +119,9 @@ def submit(self,
         else:
             dlog.debug('new task')
             self.do_submit(job_dirs, cmd, args, res, outlog=outlog, errlog=errlog)
-        time.sleep(sleep) # For preventing the crash of the tasks while submitting        
+        if 'sleep' in res and res['sleep'] > 0:
```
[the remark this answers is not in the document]
Cloudac7: So should I change the name of the very key, such as "wait_time"?

dpgen/dispatcher/LSF.py
Cloudac7: Only for the same as the code below, without any other meaning.

dpgen/dispatcher/LSF.py
Cloudac7: Thanks, it will be modified in the next commit.

dpgen/dispatcher/Batch.py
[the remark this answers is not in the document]
amcadmus: I removed the argument `sleep` from the function `submit`

PR #225, opened 2020-02-20
support plm_path that defines CV in a separate file

Manyi-Yang:


dpgen/generator/run.py
```diff
@@ -777,6 +782,8 @@ def _make_model_devi_revmat(iter_index, jdata, mdata, conf_systems):
                     plm_lines = revise_by_keys(plm_lines, rev_keys[num_lmp:], rev_item[num_lmp:])
                     with open('input.plumed', 'w') as fp:
                         fp.write(''.join(plm_lines))
+                    if use_plm_path:
```
DiracMD, line 785: copy plumed.in into the  workdir?

dpgen/generator/run.py
```diff
@@ -777,6 +782,8 @@ def _make_model_devi_revmat(iter_index, jdata, mdata, conf_systems):
...
+                       shutil.copyfile(plm_path_templ, 'plmpath.pdb')
```
DiracMD, line 786: copy the plumed.in to workdir ?

njzjz: Yes, since we have already `os.chdir(task_path)` (line 768)

PR #269, opened 2020-04-22
fix bug in DispatcherList.py. change job_record.context from list to …

dingzhaohan:
…dic.


amcadmus:
untested code. close this pr

PR #273, opened 2020-05-15
restart training if interrupted

njzjz:
thanks to Manyi Yang, who came up with this idea three months ago...

### DIFF
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -455,8 +455,9 @@ def run_train (iter_index,
         assert(train_command)
         command =  '%s train %s' % (train_command, train_input_file)
         if training_init_model:
-            command += ' --init-model old/model.ckpt'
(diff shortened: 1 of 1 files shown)


dpgen/generator/run.py
```diff
@@ -456,6 +456,7 @@ def run_train (iter_index,
         command =  '%s train %s' % (train_command, train_input_file)
         if training_init_model:
             command += ' --init-model old/model.ckpt'
+        command = "{ if [ ! -f model.ckpt.index ]; then %s; else %s --restart model.ckpt; fi }" % (command, command)
```
njzjz, line 459:
```suggestion
        if training_init_model:
            command = "{ if [ ! -f model.ckpt.index ]; then %s --init-model old/model.ckpt; else %s --restart model.ckpt; fi }" % (command, command)
        else:
            command = "{ if [ ! -f model.ckpt.index ]; then %s; else %s --restart model.ckpt; fi }" % (command, command)
```

PR #276, opened 2020-06-02
retry slurm commands when failing

njzjz:
In my experience, sometimes dpgen fails to execute slurm commands. In this commit, dpgen will sleep 60 s and retry for at most 3 times.


njzjz:
image

njzjz:
This PR "disappeared" and I am thinking a better way to resolve the problem.

AnguseZhang:
> This PR "disappeared" and I am thinking a better way to resolve the problem.

Do you have any ideas about why it disappeared? Seems  MR #313 deletes it.

image

PR #286, opened 2020-06-18
add some parameters for SSHSession

Cloudac7:
In this PR, I add some parameters for `SSHSession` from the original `paramiko.SSHClient.connect()`, including `key_filename`(local private key for ssh), `timeout` and `passphrase`(local passphrase for private key). For users' daily usage, the parameters are sometimes used but not supported by dpgen in `machine.json`, so I just add them.


Cloudac7:
Some critical bugs in this PR has been just fixed. Sorry for introducing them. And I had fixed the `time.sleep(sleep)` call, for there isn't any parameter called `sleep`, but `sleep_time` exists. After fixing, the code could work well.

PR #302, opened 2020-07-24
Fix bug if cuda_multi_tasks is used

AnguseZhang:
The train_command has been replaced in the format { xxxxx }, so if there is not a ";" before it , there will be a bug.

Like

CUDA_VISIBLE_DEVICES=2  { if [ ! -f model.ckpt.index ]; then dp train input.json; else dp train input.json --restart model.ckpt; fi }  1>> train.log 2>> train.log  &

### DIFF
--- a/dpgen/dispatcher/Batch.py
+++ b/dpgen/dispatcher/Batch.py
@@ -155,7 +155,7 @@ class Batch(object) :
             else :
                 # do not support task-wise restart
                 tmp_cmd = ' %s 1>> %s 2>> %s ' % (self.sub_script_cmd(cmd, jj, res), outlog, errlog)
-                ret += 'CUDA_VISIBLE_DEVICES=%d; %s &\n\n' % ((self.cmd_cnt % self.manual_cuda_devices), tmp_cmd)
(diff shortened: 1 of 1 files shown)


njzjz:
Will it work if we don't use `export XX=XX`?

dpgen/dispatcher/Batch.py
```diff
@@ -155,7 +155,7 @@ def _sub_script_inner(self,
             else :
                 # do not support task-wise restart
                 tmp_cmd = ' %s 1>> %s 2>> %s ' % (self.sub_script_cmd(cmd, jj, res), outlog, errlog)
-                ret += 'CUDA_VISIBLE_DEVICES=%d %s &\n\n' % ((self.cmd_cnt % self.manual_cuda_devices), tmp_cmd)
+                ret += 'CUDA_VISIBLE_DEVICES=%d; %s &\n\n' % ((self.cmd_cnt % self.manual_cuda_devices), tmp_cmd)
```
njzjz, line 158:
```suggestion
                ret += 'export CUDA_VISIBLE_DEVICES=%d; %s &\n\n' % ((self.cmd_cnt % self.manual_cuda_devices), tmp_cmd)
```
I think it won't work unless we use `export`...

AnguseZhang: Yes, I agree with you.

PR #309, opened 2020-08-16
dump job record after every submitting

njzjz:
prevent dpgen crash when submitting jobs, especially when there are many jobs to submit or task_max is set, so dpgen can restart submitting then


njzjz:
@amcadmus please review

PR #312, opened 2020-08-27
Devel

fqgong:
When I use the enhance sampling method with plumed during the exploration step of dpgen, I want to obtain the HILLS file so I can sum this file to get the free energy versus collective variables. 
If I use cp2k to do single point energy calculation at the remote server, it can't be done without basis_set and pseudo-potential files, so I need to pass these files to the remote server.


dpgen/generator/run.py
```diff
@@ -991,6 +992,8 @@ def run_model_devi (iter_index,
         forward_files += ['input.plumed']
        # backward_files += ['output.plumed']
         backward_files += ['output.plumed','COLVAR','dump.0.xyz']
```
njzjz, line 994: You don't need to use a parameter to control it. It's fine to append `HILLS` to `backward_files` directly.

PR #313, opened 2020-09-02
Add parameters to SSHSession and ignore lost atoms when necessary

Cloudac7:
Some parameters are added in order to support key phrase and key file not placed in `~/.ssh`, which is useful for safety especially when sharing the dpgen host with others. 

Also, the key `lost_ignore` is added to enable ignoring the error raised when there is lost atoms in model deviation. For some situation, for example when the model trained is poor, the phenomena may present, but it will prevent the train process going on. I just add the choice that user could choose to go on training, although the choice should be take carefully.


amcadmus:
I would not suggest adding `lost_ignore`. When the quality of the model is poor, it is recommend reduce the length of the MD simulation, rather than continue a totally wrong simulation.

Cloudac7:
> I would not suggest adding `lost_ignore`. When the quality of the model is poor, it is recommend reduce the length of the MD simulation, rather than continue a totally wrong simulation.

Well, it is true that "lost atoms" present only at the first several iterations when the model is not so good to use. At that time, it is better to shorten the sampling timescale than ignoring the error with danger. So I will draw back the part `lost_ignore`. Thanks.

PR #339, opened 2020-12-22
using "type_map" when evoking dpdata | vasp/lammps

Vibsteamer:
adding the parameter named "type_map_list" for "type_map" when using dpdata to covert config formats

### DIFF
--- a/dpgen/auto_test/lib/lammps.py
+++ b/dpgen/auto_test/lib/lammps.py
@@ -8,7 +8,7 @@ from distutils.version import LooseVersion
 from dpdata.periodic_table import Element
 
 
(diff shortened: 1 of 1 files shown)


Vibsteamer:
needs the parameter "type_map_list" defined in the pr #338 to pass the build check.

 File "/home/travis/build/deepmodeling/dpgen/dpgen/auto_test/Lammps.py", line 119, in make_input_file
    lammps.cvt_lammps_conf(os.path.join(output_dir, 'POSCAR'), os.path.join(output_dir, 'conf.lmp'))
TypeError: cvt_lammps_conf() missing 1 required positional argument: 'type_map_list'

PR #355, opened 2021-03-08
fix pymatgen v2022.0.* incompatible issue

Cloudac7:
Since last weekend, pymatgen released their v2022.0.* version, and as shown on [official document](https://pymatgen.org/#major-announcement-v2022-0), a backwards incompatible change has been introduced. I just followed the instruction official document to fix the incompatible issue for dpgen.
Also, a typo error in LSF dispatcher was fixed.


AnguseZhang:
Is this change compatible with older version of Pymatgen?

Cloudac7:
Yes, it's supported since

PR #356, opened 2021-03-08
fix pymatgen v2022.0.* incompatible issue (switch the branch)

Cloudac7:
Since last weekend, pymatgen released their v2022.0.* version, and as shown on official document, a backwards incompatible change has been introduced. I just followed the instruction official document to fix the incompatible issue for dpgen. Moreover, I checked the history code showing that the fix is supported since v2017.6.24.
Also, a typo error in LSF dispatcher was fixed.


njzjz (changes requested):
You need to add the version constraint to `setup.py`.

PR #380, opened 2021-04-26
fix "Argument list too long" for tar command

njzjz:
If the number of files is too much, it will raise an error
"Argument list too long". A solution is to split it to several
commands.

> RuntimeError: Get error code 1 in calling tar czf ...  through ssh with job: 57c454d1-1ca2-4b6b-a5c4-8c468898586e . message: /bin/bash: Argument list too long

### DIFF
--- a/dpgen/dispatcher/SSHContext.py
+++ b/dpgen/dispatcher/SSHContext.py
@@ -273,12 +273,28 @@ class SSHContext (object):
 
     def _get_files(self, 
                    files) :
-        of = self.job_uuid + '.tgz'
-        flist = ""
-        for ii in files :
-            flist += " " + ii
+        of = self.job_uuid + '.tar.gz'
         # remote tar
-        self.block_checkcall('tar czf %s %s' % (of, flist))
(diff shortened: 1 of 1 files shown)


njzjz:
@felix5572 Can I apply this patch to https://github.com/deepmodeling/dpdispatcher/blob/6ab40e219cc0676cd397a5ee124aa43aa4366caa/dpdispatcher/ssh_context.py#L389-L394?

felix5572:
Maybe another method?
Or try to wrap it a an individual function ?

https://stackoverflow.com/questions/23817787/bash-bin-tar-argument-list-too-long-when-compressing-many-files-with-tar

njzjz:
> Maybe another method?
> Or try to wrap it a an individual function ?
> 
> https://stackoverflow.com/questions/23817787/bash-bin-tar-argument-list-too-long-when-compressing-many-files-with-tar

There is a problem here: if you kill the dpgen after the tar.gz file creates, and then restart dpgen, you will get `gzip: 570aef2d-1f5f-4de3-b767-d5a2c4a8e41a.tar.gz already exists;  not overwritten`.

PR #381, opened 2021-04-27
retry slurm commands when failing

njzjz:
resubmitting #276﻿


felix5572:
I will implentment it in the deepmodeling/dpdispatcher soon

PR #385, opened 2021-04-28
remove a directory asynchronously

njzjz:
The original implementation method removes files one by one.
In some supercomputers, it's very slow to remove large numbers of files
 (e.g. directory containing trajectory) due to bad I/O performance.
Also, if the server latency is high, it is very terrible 
to execute thousands of commands synchronously.
Instead, we can call `nohup rm -rf $directory >/dev/null &`
to remove a directory asynchronously, which will save a lot 
of time in some situations.

@felix5572 said he may have a better way to implement it.

### DIFF
--- a/dpgen/dispatcher/SSHContext.py
+++ b/dpgen/dispatcher/SSHContext.py
@@ -200,9 +200,7 @@ class SSHContext (object):
 
     def clean(self) :        
         self.ssh_session.ensure_alive()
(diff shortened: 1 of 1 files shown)


amcadmus:
Some times the user dose not want a asynchronous dir removing, we should provide the option for them.

njzjz:
> Some times the user dose not want a asynchronous dir removing, we should provide the option for them.

How do we provide the option?

PR #387, opened 2021-04-30
store sftp and reduce frequency of open_sftp 

njzjz:
open_sftp and close_sftp are expensive actions, especially when the
remote server is far away from the local server. This commit stores sftp
in the object and will reduce the frequency of open_sftp.
I have tested for several days and the speed has obviously increased.
cc @felix5572


felix5572 (approved):
keep a single sftp instance is a great idea.
open and close sftp  frenquently can be very slow

Issue #386, opened 2021-04-30
[Feature Request] Check compatible version with DeePMD-kit 


AnguseZhang:
**Summary**
We should figure out which version of DeePMD-kit will be supported in DP-GEN.

**Detailed Description**

Problems occur when users specify a wrong version of "param.json" or "machine.json", such as "dp_train" and "dp train" and "input.json" of DeePMD-kit.  

I propose to deprecate the support for DeePMD-kit < 1.0, and only provide examples for DeePMD-kit >= 1.0.

Meanwhile, we're going to release DeePMD-kit 2.0. I'm wondering if there is a big difference of user interface?

AnguseZhang:
@amcadmus @njzjz  What's your opinion?

njzjz:
>  only provide examples for DeePMD-kit >= 1.0.
I agree. 
> if there is a big difference of user interface?
The training parameters have [breaking changes](https://deepmd.readthedocs.io/en/latest/train-input.html). Also, it is valuable to add `dp compress` into the workflow.

njzjz:
Note: Besides `systems`, `auto_prob_style` is also required to update.

njzjz:
Close as resolved.

PR #397, opened 2021-05-06
Decrepate support for DeePMD-kit-0.x version and refine examples and README.

AnguseZhang:
See #386


tests/generator/test_make_fp.py
```diff
@@ -810,6 +821,7 @@ def test_make_fp_pwmat(self):
         atom_types = [0, 1, 2, 2, 0, 1]
         type_map = jdata['type_map']
         _make_fake_md(0, md_descript, atom_types, type_map)
+        print("Anguse : debug pwmat ", jdata)
```
amcadmus, line 824: Remove the printing or use logging?

AnguseZhang: Fixed. I had to check this since the unittest for DP-GEN cannot pass on my MacOS, and I figured out that it's because in PWMAT interface's  unittest, a binary script https://github.com/deepmodeling/dpgen/blob/devel/tests/generator/out_data_post_fp_pwmat/02.fp/task.000.000000/poscar2config.x  was involved and could only execute on Linux.

AnguseZhang: This script was used to convert a POSCAR to atom's configuration file in PWMAT's format. I think it should be replaced with dpdata's functions.

amcadmus (approved):

examples/machine/DeePMD-kit-1.x/machine-ali.json
```diff
@@ -4,7 +4,7 @@
       "machine": {
         "batch": "shell",
         "hostname": "",
-        "password": "975481DING!",
+        "password": "PASSWORD",
```
njzjz, line 7: Maybe not related to this PR - I would suggest to use [`ssh-copy-id`](https://www.ssh.com/academy/ssh/copy-id) instead of using `password`。

njzjz (approved):

PR #420, opened 2021-05-25
support DeePMD-kit 2

### LINKED ISSUE #386: [Feature Request] Check compatible version with DeePMD-kit

njzjz:
This PR is aimmed to support DeePMD-kit 2. Fix #386.

 support deepmd 2.0
- support dp 2.0
- support dp 2.0
- support dp 2 auto_prob

### DIFF
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -319,6 +319,7 @@ def make_train (iter_index,
             raise RuntimeError('invalid setting for use_ele_temp ' + str(use_ele_temp))
     elif LooseVersion(mdata["deepmd_version"]) >= LooseVersion('2') and LooseVersion(mdata["deepmd_version"]) < LooseVersion('3'):
         # 2.x
+        jinput['training']['training_data'] = {}
         jinput['training']['training_data']['systems'] = init_data_sys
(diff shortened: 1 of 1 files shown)


AnguseZhang (approved):

AnguseZhang:
Should we provide an DeePMD-kit-2.0 example , such as CH4?

njzjz:
> Should we provide an DeePMD-kit-2.0 example , such as CH4?

Will DP parameters change any more? We can add the example when it is stable.

amcadmus:
> > Should we provide an DeePMD-kit-2.0 example , such as CH4?
> 
> Will DP parameters change any more? We can add the example when it is stable.

Basically no. We can wait until 2.0 releases.

PR #424, opened 2021-06-03
use new dpdispatcher

felix5572:
use new dpdispatcher


dpgen/data/gen.py
```diff
@@ -627,15 +653,41 @@ def run_vasp_md(jdata, mdata):
...
+    submission = Submission(
+        work_base=work_dir,
+        machine=machine,
+        resources=resources,
+        forward_common_files=forward_common_files,
+        backward_common_files=[],
+        task_list=task_list
+    )
+
+    submission.run_submission()
```
njzjz, line 689: These codes are duplicated with others. I think it's better to create a new function which can be called by every task. It will be easier to maintain.

PR #427, opened 2021-06-07
Add gromacs in model deviation

AnguseZhang:
Two things need completion in the future.
1. Use dp subcommand to replace model_devi.py
2. Provide a better DP + Gromacs user-interface.


dpgen/generator/run.py
```diff
@@ -1676,9 +1813,14 @@ def make_fp_gaussian(iter_index,
...
         fp_params = jdata['fp_params']
     cwd = os.getcwd()
+
+    model_devi_engine = jdata.get('model_devi_engine', 'lammps')
     for ii in fp_tasks:
         os.chdir(ii)
-        sys_data = dpdata.System('POSCAR').data
+        if model_devi_engine == "lammps":
+            sys_data = dpdata.System('POSCAR').data
+        elif model_devi_engine == "gromacs":
```
njzjz, line 1822: Is it designed to only support gaussian?

njzjz: By the way, do you think it's a mass here? We should try to merge the same codes of different fp styles.

dpgen/generator/run.py
```diff
@@ -1295,7 +1426,13 @@ def _make_fp_vasp_inner (modd_path,
         cwd = os.getcwd()
         for ii in fp_tasks:
             os.chdir(ii)
-            dump_to_poscar('conf.dump', 'POSCAR', type_map)
+            if model_devi_engine == "lammps":
+                dump_to_poscar('conf.dump', 'POSCAR', type_map, fmt = "lammps/dump")
+            elif model_devi_engine == "gromacs":
+                # dump_to_poscar('conf.dump', 'POSCAR', type_map, fmt = "gromacs/gro")
+                dump_to_deepmd_raw('conf.dump', 'deepmd.raw', type_map, fmt = 'gromacs/gro')
```
njzjz, line 1433: Is there any problem to use poscar instead of deepmd/raw? Or can we use deepmd/raw for lammps? It's better to use a single format for all engines, if there is no technical issues.

njzjz: BTW: I think the code here (not your PR) is also a mass.

njzjz, line 1433: This `conf.dump` is seemed to be a symbolic link of `.gromacstrj` file. Does dpdata support to read it?

dpgen/generator/run.py
```diff
@@ -62,6 +62,11 @@
 from dpgen import ROOT_PATH
 from pymatgen.io.vasp import Incar,Kpoints,Potcar
 from dpgen.auto_test.lib.vasp import make_kspacing_kpoints
+try:
+    from gromacs.fileformats.mdp import MDP
+except ImportError:
+    dlog.info("GromacsWrapper>=0.8.0 is needed for DP-GEN + Gromacs.")
```
njzjz, line 68: This information should not appear here since not everyone will install `gromacs` package.

dpgen/generator/run.py
AnguseZhang: I think it's not only supported for Gaussian.

dpgen/generator/run.py
AnguseZhang: "Is there any problem to use poscar instead of deepmd/raw?" Yes this is problem. POSCAR format does not keep the original order, (VASP doesn't care about order). But the order should be kept in Gromacs (or Amber), such as the topology can work. 
I think probably we could use DeePMD-kit's format, but all other fp methods depend on POSCAR currently, and DeepMD-kit's files should have multiple files such as type.raw, coord.raw, etc, and is not convenient for users to export and visualize. So currently I don't think clearly.

AnguseZhang: Actually,  ".gromacstrj" format  is  just a name convention. In codes, dpdata load this file by fmt gromacs/gro.

dpgen/generator/run.py
AnguseZhang: I  added GromacsWrapper dependency in setup.py, do you think it's appropriate? 
Same problems may occur on openbabel, rdkit and other softwares needed to support other engines, such as Gromacs, Amber, VASP, etc. @amcadmus  What do you think?

dpgen/generator/run.py
njzjz: Ok. We may resolve this problem later.

njzjz (approved):

PR #432, opened 2021-06-09
A place to put toymodels (doc/toymodels/)

tuoping:
1. Made directory doc which sphinx-build configurations; 
2. Made directory doc/toymodels;
3. Automatic listing of files in directory doc/toymodels by rst and sphinx-build.
Preview here:
https://dpgen-testtutorial.readthedocs.io/en/devel/

By the way, our docs are currently deployed on readthedocs. Do we want to move them to github?


doc/index.rst
```diff
@@ -0,0 +1,28 @@
+==========================
+DeePMD-kit's documentation
```
amcadmus, line 2: should be dpgen

amcadmus (changes requested):

AnguseZhang:
I have a concern. Do you intend to put models in DP-GEN's code repo?  But it will make DP-GEN's code too big, since one model could be larger than several Mb. What TF does is to create a repo for docs. https://github.com/tensorflow/docs

tuoping:
That's exactly what I mean by "move docs to github". We can open a new branch just for docs and examples.

njzjz:
Do you mean using GitHub Actions and GitHub Pages to host documents?

amcadmus:
> That's sort of what I mean by "move docs to github". We can open a new branch just for docs and examples. A repo specifically for docs is fairly common as well. However, our community is relatively diverse, which means double repo for each project, one for code and another for docs. I don't think that's a good idea.

Could you plz explain the benefit of using a separate repo for doc?

njzjz (approved):

tuoping:
> Do you mean using GitHub Actions and GitHub Pages to host documents?

Yes.


> > That's sort of what I mean by "move docs to github". We can open a new branch just for docs and examples. A repo specifically for docs is fairly common as well. However, our community is relatively diverse, which means double repo for each project, one for code and another for docs. I don't think that's a good idea.
> 
> Could you plz explain the benefit of using a separate repo for doc?

Actually, I don't think a separate repo for doc is a good idea. A separate branch is more practical. The benefit of a separate branch is more centralized DNS management. Although it's not necessary, we can simply change the domain name of readthedocs page.

tuoping:
> Do you mean using GitHub Actions and GitHub Pages to host documents?

Yes.


> > That's sort of what I mean by "move docs to github". We can open a new branch just for docs and examples. A repo specifically for docs is fairly common as well. However, our community is relatively diverse, which means double repo for each project, one for code and another for docs. I don't think that's a good idea.
> 
> Could you plz explain the benefit of using a separate repo for doc?

Actually, I don't think a separate repo for doc is a good idea. A separate branch is more practical. The benefit of a separate branch is more centralized DNS management. Although it's not necessary, we can simply change the domain name of readthedocs page.

tuoping:
After a brief discussion with tansongchen. We think it's more convenient to use readthedocs. So that we don't have to maintain a workflow for doc.

njzjz:
readthedocs supports to add a project into another one as a subproject, and they will share the same domain. Do you think it's a good idea to make dpgen as a sub-project as deepmd-kit (or we create a new one as the main project)?

tuoping:
I vote "create a new one".

amcadmus:
> 
> 
> readthedocs supports to add a project into another one as a subproject, and they will share the same domain. Do you think it's a good idea to make dpgen as a sub-project as deepmd-kit (or we create a new one as the main project)?

dpgen is not a subject of deepmd-kit

amcadmus:
Now the RTD page is blank. Is this PR a preparation for a serious construction of the RTD page?

njzjz:
How about creating a new repository called `docs`, which only contains links to other documents or something else?

@tuoping Yesterday I gave you the access to the current documents.

tuoping:
According to the document of RTD. The first step to set up custom domain seems to require a change in the DNS of our home page "https://deepmodeling.org/".@njzjz

In order to setup your custom domain, follow these steps:
1. For a subdomain like docs.example.com, add a CNAME record in your DNS that points the domain to readthedocs.io. For a root domain like example.com use an ANAME or ALIAS record pointing to readthedocs.io.
2. ……

tuoping:
I have built a test page here https://dpgen-testtutorial.readthedocs.io/en/devel/
@amcadmus If you think it works, we can go with this.

njzjz:
> According to the document of RTD. The first step to set up custom domain seems to require a change in the DNS of our home page "https://deepmodeling.org/". @njzjz 
> 
> 
> 
> In order to setup your custom domain, follow these steps:
> 
> 1. For a subdomain like docs.example.com, add a CNAME record in your DNS that points the domain to readthedocs.io. For a root domain like example.com use an ANAME or ALIAS record pointing to readthedocs.io.
> 
> 2. ……

You can use docs.deepmodeling.org.

tuoping:
Is the CNAME record added in deepmodeling.org for docs.deepmodeling.org? Because it's not working. The custom domain name docs.deepmodeling.org is configured on readthedocs. But docs.deepmodeling.org is still 404.

njzjz:
> Is the CNAME record added in deepmodeling.org for docs.deepmodeling.org? Because it's not working. The custom domain name docs.deepmodeling.org is configured on readthedocs. But docs.deepmodeling.org is still 404.

This domain is not maintained by me... I think it's @tansongchen

PR #433, opened 2021-06-09
added ABACUS interface for first principle calculations

Liu-RX:
Added ABACUS plane wave self consistent field (pescf) interface for first principle calculations in Run process.


dpgen/generator/run.py
```diff
@@ -1771,6 +1772,41 @@ def make_fp_pwscf(iter_index,
...
+    fp_tasks = _make_fp_vasp_configs(iter_index, jdata)
+    if len(fp_tasks) == 0 :
+        return
+    # make abacus/pw/scf input
+    iter_name = make_iter_name(iter_index)
+    work_path = os.path.join(iter_name, fp_name)
+    fp_pp_files = jdata['fp_pp_files']
+    if 'user_fp_params' in jdata.keys() :
+        fp_params = jdata['user_fp_params']
+        #user_input = True
```
AnguseZhang, line 1787: Are these codes similar to `make_fp_abacus_pw` in `relabel.py`? Do you consider merging them?

dpgen/tools/relabel.py
```diff
@@ -131,6 +147,10 @@ def create_init_tasks(target_folder, param_file, output, fp_json, verbose = True
                     fp_params = fp_jdata['fp_params']
                     user_input = False
                 make_pwscf('.', fp_params, mass_map, fp_pp_files, fp_pp_files, user_input)
+            elif fp_style == "abacus/scf":
+                if "user_fp_params" not in fp_jdata:
+                    raise RuntimeError("'user_fp_params' key and its value are necessary for 'abacus/scf' fp_style.")
+                make_abacus_pw_scf(".", fp_params, mass_map, fp_pp_files, fp_pp_files)
             elif fp_style == 'siesta':
```
AnguseZhang, line 154: Did you try `relabel.py` with Abacus?  I didn't see the unit tests for this part.

AnguseZhang (changes requested):
About the main code, I've commented in my review.
I think there needs changes on the unit tests.  Currently there're more than 600k lines added, but most of them are files for unit tests, and I think most of them can also be saved. Such as you don't need a full file of pseudo-potential file (POTCAR, UPF,etc.), but instead just need a link or an empty file. (I didn't see you test the information inside them). 
Also, many POSCARS are repeated. PLEASE divide the tests into as smaller as you can, otherwise the codes will become more and more heavy and unmaintainable.

dpgen/generator/run.py
Liu-RX: I cannot do this. Actually I did not thoroughly read the Run process workflow, so I just wrote the interface imitating the pwscf interface. And I found that there are two sets of functions making input files in run.py and relabel.py for every first principle software, such as `make_fp_siesta` and `make_siesta`, `make_fp_pwscf` and `makepwscf`. I followed these writing mode.

amcadmus (commented):
What is the reason of modifying test data files like  tests/generator/out_data_nbands_esti/md.000300K/POSCAR ?

Liu-RX:
> What is the reason of modifying test data files like tests/generator/out_data_nbands_esti/md.000300K/POSCAR ?

I rerun the test codes on my work station. The file is replaced with a new same on in the process, and GitHub denote is as a modification. I will redo the modification.

Liu-RX:
The reason of the run test failure is that I cited ABACUS interface in dpdata I added the other days. The test package has not yet included the dpdata interface addition, thereby triggering the test error. Should I include the dpdata ABACUS interface part in dpgen to make it run successfully?

AnguseZhang (commented):
> The reason of the run test failure is that I cited ABACUS interface in dpdata I added the other days. The test package has not yet included the dpdata interface addition, thereby triggering the test error. Should I include the dpdata ABACUS interface part in dpgen to make it run successfully?

@amcadmus  Shall we release a new version of dpdata such that new functions can be used by DP-GEN?

dpgen/generator/run.py
AnguseZhang: @amcadmus Are we stilling use "relabel.py"? Do you think it necessary to add unit tests for this function?

tests/generator/param-pyridine-abacus.json
```diff
@@ -0,0 +1,117 @@
...
+    "init_data_sys":	["Pyridine-I",
+			 "Pyridine-II"
+			],
+    "init_batch_size":	[1, 1],
+    "sys_configs":	[
+	["/home/linfengz/SCR/wanghan/data/pyridine/pyI.POSCAR.01x01x01/01.scale_pert/sys-0080-0080-0016/scale-1.000/00009?/POSCAR"],
+	["/home/linfengz/SCR/wanghan/data/pyridine/pyI.POSCAR.01x01x01/01.scale_pert/sys-0080-0080-0016/scale-1.000/0000[7-8]?/POSCAR"],
+	["/home/linfengz/SCR/wanghan/data/pyridine/pyI.POSCAR.01x01x01/01.scale_pert/sys-0080-0080-0016/scale-1.000/0000[5-6]?/POSCAR"],
+	["/home/linfengz/SCR/wanghan/data/pyridine/pyI.POSCAR.01x01x01/01.scale_pert/sys-0080-0080-0016/scale-1.000/0000[0-4]?/POSCAR"],
+	["/home/linfengz/SCR/wanghan/data/pyridine/pyII.POSCAR.01x01x01/01.scale_pert/sys-0080-0080-0016/scale-1.000/00009?/POSCAR"],
```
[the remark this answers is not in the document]
Liu-RX: Sorry I did not... I copied the json file just for convenience in the first place, so long as it just serve as a test for input script. I will replace the case with the methane one.

dpgen/generator/run.py
Liu-RX: > That's my point. "relabel.py" serves to re-calculate the first-principles jobs if there is need. Actually I haven't use this fucntion, either. If you cannot use that function and make sure that your modification works, you should't write codes based on it. Just go ahead.

Let me try the work flow out without the codes in "relabel.py". If it works, I will remove these codes.

Liu-RX:
I made the changes as suggested by @AnguseZhang . I used the methane case throughout the whole test process, and removed the codes added in "relabel.py" which are not used.

AnguseZhang (approved):

dpgen/generator/run.py
amcadmus: Yes we are using relabel.py for relabeling a system under different fp settings.

PR #440, opened 2021-06-21
add amber model_devi_engine and amber/corr fp_style

njzjz:
~(and make the code more like a 💩 hill)~

TODO: add an unittest

### DIFF
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -1276,7 +1276,7 @@ def run_model_devi (iter_index,
         backward_files = ["%s.tpr" % deffnm, "%s.log" %deffnm , 'model_devi.out', 'model_devi.log']
     elif model_devi_engine == "amber":
         # TODO: currently all options are written in lmp_exec, but we may move them here
-        command = [mdata['lmp_exec']]
+        commands = [lmp_exec + " -O -p qmmm.parm7 -c init.rst7 -i init.mdin -o rc.mdout -r rc.rst7 -x rc.nc -inf rc.mdinfo"]
         forward_files = ['init.rst7', 'init.mdin', 'qmmm.parm7', 'TEMPLATE.disang']
         backward_files = ['rc.mdout', 'rc.nc', 'rc.rst7', 'rc.mdinfo', 'TEMPLATE.dumpave']
 
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -787,6 +787,12 @@ def make_model_devi (iter_index,
                 system.to_lammps_lmp(os.path.join(conf_path, lmp_name))
             elif model_devi_engine == "gromacs":
                 pass
+            elif model_devi_engine == "amber":
+                # Jinzhe's specific Amber version
+                conf_name = make_model_devi_conf_name(sys_idx[sys_counter], conf_counter)
+                rst7_name = conf_name + '.rst7'
+                # link restart file
+                os.symlink(cc, os.path.join(conf_path, rst7_name))
             conf_counter += 1
         sys_counter += 1
 
@@ -800,6 +806,8 @@ def make_model_devi (iter_index,
             _make_model_devi_native(iter_index, jdata, mdata, conf_systems)
         elif model_devi_engine == "gromacs":
             _make_model_devi_native_gromacs(iter_index, jdata, mdata, conf_systems)
+        elif model_devi_engine == "amber":
+            _make_model_devi_amber(iter_index, jdata, mdata, conf_systems)
         else:
             raise RuntimeError("unknown model_devi engine", model_devi_engine)
     elif input_mode == "revise_template":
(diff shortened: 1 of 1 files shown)


AnguseZhang:
Got. I may review this on Wed/Thur.

AnguseZhang (commented):
My concern is that both this commit for Amber and my last commit for Gromacs are so specific that no one else can use these functions. Do you think that we should support more general cases, just like CH4, and make this as a milestone for a PR?

njzjz:
> My concern is that both this commit for Amber and my last commit for Gromacs are so specific that no one else can use these functions. Do you think that we should support more general cases, just like CH4, and make this as a milestone for a PR?

What I suggest is to make DP-GEN support plugins... but it may take a lot of effort. If someone else also has this idea, we can implement it together.

AnguseZhang:
> > My concern is that both this commit for Amber and my last commit for Gromacs are so specific that no one else can use these functions. Do you think that we should support more general cases, just like CH4, and make this as a milestone for a PR?
> 
> What I suggest is to make DP-GEN support plugins... but it may take a lot of effort. If someone else also has this idea, we can implement it together.

What do you meany by supporting plugins? Is there an example? I don't think plugin like plumed is similar to DP-GEN.

njzjz:
> > > My concern is that both this commit for Amber and my last commit for Gromacs are so specific that no one else can use these functions. Do you think that we should support more general cases, just like CH4, and make this as a milestone for a PR?
> > 
> > 
> > What I suggest is to make DP-GEN support plugins... but it may take a lot of effort. If someone else also has this idea, we can implement it together.
> 
> What do you meany by supporting plugins? Is there an example? I don't think plugin like plumed is similar to DP-GEN.

Like LAMMPS? In a LAMMPS plugin, one can add custom `fix` and `pair` styles by define a new class (e.g. [deepmd's LAMMPS plugin](https://github.com/deepmodeling/deepmd-kit/blob/afc557abcf17eb7d37f2bf93624fc060122dc241/source/lmp/pair_deepmd.cpp#L232-L233)). For DP-GEN, `model_devi` and `fp` styles can also be customize.

AnguseZhang:
> > > > My concern is that both this commit for Amber and my last commit for Gromacs are so specific that no one else can use these functions. Do you think that we should support more general cases, just like CH4, and make this as a milestone for a PR?
> > > 
> > > 
> > > What I suggest is to make DP-GEN support plugins... but it may take a lot of effort. If someone else also has this idea, we can implement it together.
> > 
> > 
> > What do you meany by supporting plugins? Is there an example? I don't think plugin like plumed is similar to DP-GEN.
> 
> Like LAMMPS? In a LAMMPS plugin, one can add custom `fix` and `pair` styles by define a new class (e.g. [deepmd's LAMMPS plugin](https://github.com/deepmodeling/deepmd-kit/blob/afc557abcf17eb7d37f2bf93624fc060122dc241/source/lmp/pair_deepmd.cpp#L232-L233)). In DP-GEN, `model_devi` and `fp` styles can also be customize.

This is gonna like refacting DP-GEN codes,  I talked with @amcadmus and came to conclusion that it's necessary but we don't have time for now.  Ok let's go back to this PR. I try to be more specfic.  Recall that when previously other developers implement a new fp style, like PWMAT, SIESTA, CP2K, e.t.c, their milestone is to use DP-GEN to run a complete case where fp calculation method is replaced wtih the new method. So I think if we add a new model_devi engine,like Gromacs and Amber, the same milestone should be applied. Otherwise, we cannot tell someone else who is interested in trying DP-GEN with Gromacs and Amber  how to use.

njzjz:
Do you mean it should be possible to mix any `model_devi_engine` and any `fp_style`? Err I think it's a bit hard to do so if we don't refactor codes, as there is too much thing coupling here. For example, the output of `model_devi` is used to start `fp_style`, and we should first define an unified input and output for decoupling.

njzjz:
In my idea, currently a part of `make_fp` is actually the post-processing of `model_devi` and needs to move to `post_model_devi`. 
https://github.com/deepmodeling/dpgen/blob/e89d7fc67e59bccadfc93cc3f12d8b205333d6fc/dpgen/generator/run.py#L1272

And also, it's very strange here to switch `model_devi_engine` in a `fp` function. The output of `model_devi` should be unified to be used by `fp`.
https://github.com/deepmodeling/dpgen/blob/e89d7fc67e59bccadfc93cc3f12d8b205333d6fc/dpgen/generator/run.py#L1878-L1883

AnguseZhang:
> Do you mean it should be possible to mix any `model_devi_engine` and any `fp_style`? Err I think it's a bit hard to do so if we don't refactor codes, as there is too much thing coupling here. For example, the output of `model_devi` is used to start `fp_style`, and we should first define an unified input and output for decoupling.

I think this is the second stage. The first stage is to successfully run an easy and general case in Amber or Gromacs, and provide examples `param.json`.

AnguseZhang:
> In my idea, currently a part of `make_fp` is actually the post-processing of `model_devi` and needs to move to `post_model_devi`.
> https://github.com/deepmodeling/dpgen/blob/e89d7fc67e59bccadfc93cc3f12d8b205333d6fc/dpgen/generator/run.py#L1272
> 
> And also, it's very strange here to switch `model_devi_engine` in a `fp` function. The output of `model_devi` should be unified to be used by `fp`.
> https://github.com/deepmodeling/dpgen/blob/e89d7fc67e59bccadfc93cc3f12d8b205333d6fc/dpgen/generator/run.py#L1878-L1883

I argee.

Issue #453, opened 2021-07-03
[Feature Request] _different trust level for different sys_configs_ 


changxiaoju:
**Summary**

when selecting the structures of 01.model_devi, different sys-configs may have differnet model deviation, and it is hard to set trust levels in this case cause only sys-configs with middle model deviation can be well sampled. And it is expensive to explore just one sys-config every iteration.

**Detailed Description**

i wish dpgen can have multi-setting of trust level for different sys_configs. for example,  the type of `"model_devi_f_trust_lo" `and`"model_devi_f_trust_hi"` can be a list of float, and each number of the list represent the relative value for different `"sys_idx"`.

**Further Information, Files, and Links**

Or, if dpgen can explore along concentration and component might also solve my problem.

PR #461, opened 2021-07-10
Update run.py | allow assigning different  activation functions to em…

Vibsteamer:
…bedding nets and fitting nets of each model

Scenario:  In some circumstances, models trained with the same assignment of the kinds of activation functions may give close inferences for the same configuration that is sampled from 01.model_devi, even when the feature in atomic environment of this configuration has not been well covered by the current data sets (models give similarly bad inference).
The degree of the “similarly bad” maybe related to the feature in atomic environment, then may be different when configurations are sampled from different phases.

Problem:  In some circumstances, the deviation of the model ensemble may be scattered in a wide range when 01.model_devi is initiated from different phases, thus introduce difficulties in setting the common trust_lo. Similar description could be found in issue #453.

This update:  Further update the expected format of the value for the key "model_devi_activation_func" in param.json, from a list (i.e. ["tanh","tanh","gelu","gelu"] for 4 models) into a list of list (i.e. [["tanh","tanh"],["tanh","gelu"],["gelu","tanh"],["gelu","gelu"]] for 4 models ). The index of the second dimension allows assigning different activation functions to embedding nets and fitting nets of the same model (The original version already allows assigning different activation functions to different models, but within each model, embedding and fitting nets using the same ones).  

note: This update is a preview feature. Please ensure this feature is secure and applicable for your occasion. Large enough "stop_batch" might be needed to avoid models with different activation functions to deviate on configurations that had been well covered by the current data sets, due to their different sensitivities to training lengths thus possibly insufficient training of particular models. The function of "init-model" supported by DP-GEN maybe a good choice to try in some circumstances (please see related keys such as "training_reuse_iter" and "training_init_model").


AnguseZhang:
Ok

---Original---
From: "Han ***@***.***&gt;
Date: Tue, Jul 13, 2021 14:27 PM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;"Review ***@***.***&gt;;
Subject: Re: [deepmodeling/dpgen] Update run.py | allow assigning different  activation functions to em… (#461)




 
@amcadmus requested your review on: #461 Update run.py | allow assigning different  activation functions to em….

AnguseZhang (changes requested):
Thanks for your contribution!
1. This should be not a breaking change to the old behaviors. I suggest that add a detection whether users indeed to separate the activation functions in embedding network and fitting net. If users only provide a list of strings  ["tanh", "gelu", "tanh", "gelu"]  (which is the old behavior), instead of a list of list of strings [ ["tanh", "tanh"], ["gelu", "gelu"]], you should allow this old setting by detecting the structure of `model_devi_activation_func` and decide what to do according to the input.he
2.  Could you please complete the documentation related to this part? Such as README.md 
3.  I also wonder the performances of this protocol. Are there some tests? What's the performance?

amcadmus (approved):
1, I recommend to make this pr backward compatible, i.e. when `[ 'gelu', 'tanh']` is input, it is converted to `[ ['gelu', 'gelu'], ['tanh', 'tanh'] ]`. 
2, the document should be updated.

amcadmus (changes requested):
1, I recommend to make this pr backward compatible, i.e. when [ 'gelu', 'tanh'] is input, it is converted to [ ['gelu', 'gelu'], ['tanh', 'tanh'] ].
2, the document should be updated.

Vibsteamer:
> Thanks for your contribution!
> 
> 1. This should be not a breaking change to the old behaviors. I suggest that add a detection whether users indeed to separate the activation functions in embedding network and fitting net. If users only provide a list of strings  ["tanh", "gelu", "tanh", "gelu"]  (which is the old behavior), instead of a list of list of strings [ ["tanh", "tanh"], ["gelu", "gelu"]], you should allow this old setting by detecting the structure of `model_devi_activation_func` and decide what to do according to the input.he
> 2. Could you please complete the documentation related to this part? Such as README.md
> 3. I also wonder the performances of this protocol. Are there some tests? What's the performance?
> 
> Yuzhi

3. There are only results comparing the default setting and net-resolved setting (surface geometry of pure metal ):  
fa40995636b00bbe06271479c380a08
e057cc3ca7ba3fd28cfb0cbf3486c75

Vibsteamer:
> 1, I recommend to make this pr backward compatible, i.e. when [ 'gelu', 'tanh'] is input, it is converted to [ ['gelu', 'gelu'], ['tanh', 'tanh'] ].
> 2, the document should be updated.

Pleas see updates.

Vibsteamer:
WARNING: the effect of this feature has not be testified by any published works (to my knowledge), should the note be added to README.md?

Namely, the last para of the original commnet:
"note: This update is a preview feature. Please ensure this feature is secure and applicable for your occasion. Large enough "stop_batch" might be needed to avoid models with different activation functions to deviate on configurations that had been well covered by the current data sets, due to their different sensitivities to training lengths thus possibly insufficient training of particular models. The function of "init-model" supported by DP-GEN maybe a good choice to try in some circumstances (please see related keys such as "training_reuse_iter" and "training_init_model")."

amcadmus (approved):

PR #463, opened 2021-07-12
Fix parameter error for LSF GPU tasks

Cloudac7:
For GPU tasks , the positions of two parameters `task_per_node` and `numb_gpu` were written wrongly, causing tasks in PEND state. So here they are corrected. 

Also, the support of `j_exclusive` is added to enable feature that allow a task to run on a card without other tasks submitted even with enough CPU cores available in the queue.


AnguseZhang (approved):
Could you please add the documentation in README ?

njzjz (approved):

amcadmus:
@Cloudac7 an update of the doc is required.

Cloudac7:
> @Cloudac7 an update of the doc is required.

OK. I will give a introduction in the doc.

Cloudac7:
Documentation for added parameters has been added.

PR #464, opened 2021-07-15
fix NSW value in INCAR; fix local_root check

felix5572:
fix NSW value in INCAR; fix local_root check


amcadmus (approved):

examples/init/INCAR_methane.rlx
```diff
@@ -12,7 +12,7 @@ ISIF=2
 ISMEAR=0
 SIGMA=0.20000
 IBRION=-1
-NSW=10
+NSW=0
```
amcadmus, line 15: Why setting NSW=0 in a relaxation INCAR? With NSW=0 the configuration will not be relaxed.

amcadmus (changes requested):

felix5572:
Oh, I mistook it as single frame calculation.
And the parameter NSW  will only repeat the single frame calculation many times when IBRION=-1.


in vasp docs:
Description: IBRION determines how the ions are updated and moved.
IBRION=-1: no update.
IBRION=2: ionic relaxation (conjugate gradient algorithm).

amcadmus (approved):

PR #468, opened 2021-07-17
fix bug in example INCAR_methane.md; will run a MD

felix5572:
fix bug in example INCAR_methane.md; will run a MD


felix5572:
@amcadmus

PR #469, opened 2021-07-19
refactor model_devi

njzjz:
This PR has not been finished. However, through this draft PR, it will be easy to see what is changed.

### DIFF
--- a/dpgen/generator/model_devi.py
+++ b/dpgen/generator/model_devi.py
@@ -92,10 +92,13 @@ class Trajectory(ABC):
     
     Parameters
     ----------
+    engien: ModelDeviEngien
+        Engien
     directory: str
         The directory of a simulation
     """
-    def __init__(self, directory: str) -> None:
+    def __init__(self, engien: ModelDeviEngien, directory: str) -> None:
+        self.engien = engien
         self.directory = directory
         self.status = 0
     
@@ -118,20 +121,20 @@ class Trajectory(ABC):
         raise NotImplementedError("Not implemented")
     
     @abstractmethod
-    def get_frames(self, idx: List[Tuple[int]]) -> List["Frame"]:
-        """Get list of frames from idx.
+    def get_frame(self, idx: Tuple[int]) -> "Frame":
+        """Get a frame from idx.
 
         Parameters
         ----------
-        idx: list[tuple[int]]
-            List of indexes.
+        idx: tuple[int]
+            Index.
         
         Returns
         -------
-        frames: list[Frame]
-            List of frames.
+        frames: Frame
+            Frames.
         """
-        return [Frame(self, ii) for ii in idx]
+        return Frame(self, idx)
     
     def open_trajectory(self):
         """Open the trajectory.
--- a/dpgen/generator/model_devi.py
+++ b/dpgen/generator/model_devi.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 from abc import ABC, abstractmethod
-from typing import Iterator, List, TYPE_CHECKING, Tuple, Iterator
+from typing import Iterator, List, TYPE_CHECKING, Tuple, Iterator, Dict
 from dpgen.plugin import Plugin
 
 if TYPE_CHECKING:
@@ -53,7 +53,7 @@ class ModelDeviEngien(ABC):
         raise NotImplementedError("Not implemented")
 
     @abstractmethod
-    def get_running_parameters(self, work_path: str) -> Tuple[str, List[str]]:
+    def get_running_parameters(self, work_path: str) -> Dict[str]:
         """Get running parameters.
 
         Parameters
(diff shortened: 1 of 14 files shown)


dpgen/generator/styles/gromacs.py
```diff
@@ -0,0 +1,125 @@
...
+try:
+    from gromacs.fileformats.mdp import MDP
+except ImportError:
+    # JZ: Do not excuse users using other eigiens
+    #dlog.info("GromacsWrapper>=0.8.0 is needed for DP-GEN + Gromacs.")
+    pass
+
+
+@ModelDeviEngien.register("gromacs")
+class LAMMPSEngien(ModelDeviEngien):
```
felix5572, line 18: Could you please clarify what is the relation of  decorator register('gromcas') and LAMMPSEngien here.

njzjz: It's a typo, though it doesn't matter.

dpgen/generator/model_devi.py
```diff
@@ -0,0 +1,191 @@
...
+            generate the working directory to run simulations
+        conf_name: str
+            file name of initial system
+        models: list[str]
+            list of models
+        """
+        raise NotImplementedError("Not implemented")
+
+    @abstractmethod
+    def get_running_parameters(self, work_path: str) -> Tuple[str, List[str]]:
```
felix5572, line 56: the type of return value might be `dict`

dpgen/generator/model_devi.py
```diff
@@ -0,0 +1,191 @@
...
+        
+        Notes
+        -----
+        The index of model deviation should be as the same as frames.
+        """
+        # for performance, returns a numpy array instead of a list
+        raise NotImplementedError("Not implemented")
+    
+    @abstractmethod
+    def get_frame(self, idx: Tuple[int]) -> "Frame":
```
felix5572, line 124: get single frame. so `idx:int` be `tuple` ?

dpgen/generator/model_devi.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+        self.status = 1
+
+    def close_trajectory(self):
+        """Close the trajectory. Call this method when all frames have
+        been read.
+        """
+        self.status = 0
+    
+    def __lt__(self, other: "Trajectory") -> bool:
```
felix5572, line 155: Is it a good behaviour to compare two strings？

dpgen/generator/styles/gromacs.py
```diff
@@ -0,0 +1,125 @@
...
+    from gromacs.fileformats.mdp import MDP
+except ImportError:
+    # JZ: Do not excuse users using other eigiens
+    #dlog.info("GromacsWrapper>=0.8.0 is needed for DP-GEN + Gromacs.")
+    pass
+
+
+@ModelDeviEngien.register("gromacs")
+class GromacsEngien(ModelDeviEngien):
+    def make_input(self, iter_index:int, sys_index:int, directory:Iterator[str], conf_name: str, models: List[str]):
```
felix5572, line 19: If `directory` is an `Iterator`, use `directories` could be better?

dpgen/generator/styles/gromacs.py
```diff
@@ -0,0 +1,125 @@
...
+class GromacsFrame(Frame):
+    def read_frame(self) -> dpdata.System:
+        fidx = self.idx[0]
+        time = self.trajectory.time[fidx]
+        conf_name = os.path.join(self.trajectory.directory, 'traj', '%d.gromacstrj' % time)
+        type_map = self.trajectory.engien.jdata['type_map']
+        system = dpdata.System(conf_name, type_map=type_map, fmt='gromacs/gro')
+        return system
+
+def _make_model_devi_native_gromacs(iter_index, task_paths, jdata, mdata, cc, models):
```
felix5572, line 68: Maybe this could be a function of class GromacsEngien instead of an individual function?

dpgen/generator/model_devi.py
njzjz: Fixed.

dpgen/generator/model_devi.py
njzjz: Fixed.
A new `Trajectory` should implement thier own `Frame`, as the current `Frame` do nothing in `read_frame` method. So this should be an abstractmethod.

dpgen/generator/model_devi.py
njzjz: The idea of comparing is let frames with the same trajectory open together, considering some types of trajectories have only one file to read all frames. So, the way to sort `Trajectory` is not important. It's just used to sort `Frame`:
https://github.com/deepmodeling/dpgen/blob/11a77abba134d710ef5881c6a19da6f3e0972a2d/dpgen/generator/model_devi.py#L187-L191

dpgen/generator/styles/gromacs.py
njzjz: True.

dpgen/generator/styles/gromacs.py
njzjz: I can update the `GromacsEngien` in this PR. But the work for `LAMMPSEngien` is too heavy. I hope someone could help me, maybe in a separate PR.

dpgen/generator/styles/gromacs.py
```diff
@@ -0,0 +1,130 @@
...
+    from gromacs.fileformats.mdp import MDP
+except ImportError:
+    # JZ: Do not excuse users using other eigiens
+    #dlog.info("GromacsWrapper>=0.8.0 is needed for DP-GEN + Gromacs.")
+    pass
+
+
+@ModelDeviEngien.register("gromacs")
+class GromacsEngien(ModelDeviEngien):
+    def make_input(self, iter_index: int, sys_index: int, directories: Iterator[str], conf_name: str, models: List[str]):
```
njzjz, line 19: @AnguseZhang I think the Gromacs's logic here is a bit strange that is different from LAMMPS. LAMMPS can convert any format to `lammps/lmp` through dpdata, but Gromacs just copies directories. Could dpdata support Gromacs as well? Should we keep difference between two engiens here?

PR #475, opened 2021-07-21
Add .github/workflow/mirror_to_gitee.yml

tuoping:
I found the problem. I've been invoking action in my personal repository: github.com/tuoping/dpgen. So the ssh-private-key in organization deepmodeling cannot be visited. We have to test the action after PR.


tuoping:
In addition, may I ask how did you enable to sync issue from gitee to github?

.github/workflows/mirror_gitee.yml
```diff
@@ -15,5 +15,5 @@ jobs:
           ORGANIZATION: deepmodeling
           SSH_PRIVATE_KEY: ${{ secrets.SYNC_GITEE_PRIVATE_KEY }}
         with:
-          source-repo: "git@github.com:deepmodeling/dpgen.git"
+          source-repo: "http://github.com/deepmodeling/dpgen.git"
```
njzjz, line 18: use https instead of http
```suggestion
          source-repo: "https://github.com/deepmodeling/dpgen.git"
```

njzjz (changes requested):

njzjz:
> In addition, may I ask how did you enable to sync issue from gitee to github?

A simple method... Set up [this bot](https://fire.fundersclub.com) and then forward the Gitee notification emails to it.

tuoping:
Sorry about that. But this action still failed with the same error message in my repository tuoping/dpgen. I still think the action need to be invoked in deepmodeling/dpgen.

njzjz:
Did you set the secret in your own repo?

tuoping:
The point of this is to test the secret in organization. That's not a bad idea though, to set the secret in my own repo and PR with the mirror_to_gitee.yml file. This way the secret in organization will be of no use.

njzjz:
If you want to test the secret in organization, you need to push the branch into the organization repository. (or ask someone who has the access to help you)

PR #482, opened 2021-07-26
Allow user defined forward_common_files and backward_files

AnguseZhang:
Allow user defined forward_common_files and backward_files
Detailed Description

In vasp, users need to add a file vdw_kernel.bindat as input files.
In plumed, users need to collect HILLS. See Devel #312

Other updates:
1. I also clean decide_machine and unify the converting interface. For now, we would allow only one machine. 
2. "lmp_command" is deprecated, we will adopt "model_devi_command".
3. Update some docs.


dpgen/generator/run.py
```diff
@@ -2158,10 +2169,10 @@ def run_fp (iter_index,
             mdata) :
     fp_style = jdata['fp_style']
     fp_pp_files = jdata['fp_pp_files']
-
+    backward_files  = ['fp.log']
```
njzjz, line 2172: Note that only VASP uses the default `log_file`. Others set it to `output`.
https://github.com/deepmodeling/dpgen/blob/d22316e71a4ae0b2cd8246b83771931e2bd778e4/dpgen/generator/run.py#L2190

njzjz (changes requested):

dpgen/generator/run.py
```diff
@@ -1227,6 +1227,9 @@ def run_model_devi (iter_index,
 
     cwd = os.getcwd()
 
+    user_forward_files = mdata.get("model_devi" + "_user_forward_files", [])
+    forward_files += [os.path.basename(file) for file in user_forward_files]
+    backward_files += mdata.get("model_devi" + "_user_backward_files", [])
```
amcadmus, line 1232: the convention of xxx_user_forward_files, where xxx being model_devi, train, fp, is not explained in README.

dpgen/remote/decide_machine.py
```diff
@@ -285,4 +319,4 @@ def decide_fp_machine(mdata):
 					profile['command'] = mdata['fp_command']
 					json.dump(profile, _outfile, indent = 4)
 	return mdata
-
+'''
```
amcadmus, line 322: What is this?

examples/CH4-refact-dpdispatcher/machine-ali-ehpc.json
```diff
@@ -46,7 +46,11 @@
         "queue_name": "T4_4_15",
         "group_size": 5,
         "source_list": ["/home/fengbo/deepmd.1.2.4.env"]
-      }
+      },
+      "_comments" : "In user_forward_files, define input files to be uploaded.",
+      "user_forward_files" : [],
+      "_comments" : "In user_backward_files, define output files to be collected.",
+      "user_backward_files" : ["HILLS"]
```
amcadmus, line 53: Shouldn't it be model_devi_user_backward_files?

examples/init/INCAR_methane.rlx
```diff
@@ -11,8 +11,8 @@ NELMIN=4
 ISIF=2
 ISMEAR=0
 SIGMA=0.20000
-IBRION=2
-NSW=50
+IBRION=-1
+NSW=0
```
amcadmus, line 15: Why do you introduce the change here?

amcadmus (changes requested):
We should distinguish `user_forward_files` and `user_forward_common_files`. The former the task specific and the latter is common and symbol linked to all tasks. This convention is consistent to `forward_files` and `forward_common_files`. 

I also think we do not need to implement a new `symlink_user_forward_files` for user defined forward files. Just put all files need to be linked to `user_forward_common_files`.

dpgen/generator/run.py
AnguseZhang: Done

dpgen/generator/run.py
AnguseZhang: The directories of forward_common_files and forward_files are different. 
Take an example. If you ls 
```
JOBDIR/forward_common_file
JOBDIR/task.000.000/forward_file
JOBDIR/task.000.000/forward_file
```

Forward_common_files will be not put into the directories of tasks, this will cause the inconvenience for users. 
Actually we only  forward_common_files in two places: the model paths in training data set https://github.com/deepmodeling/dpgen/blob/da494650314b3eebcf31ff20a7c195b309cb1a6a/dpgen/generator/run.py#L573 and in model_devi https://github.com/deepmodeling/dpgen/blob/da494650314b3eebcf31ff20a7c195b309cb1a6a/dpgen/generator/run.py#L1195 . Even in processing POTCAR , we treated it as forward_files. https://github.com/deepmodeling/dpgen/blob/da494650314b3eebcf31ff20a7c195b309cb1a6a/dpgen/generator/run.py#L2163
 So that's why I create a new function `symlink_user_forward_files`

dpgen/remote/decide_machine.py
AnguseZhang: Comment.  I explained the reason in https://github.com/deepmodeling/dpgen/blob/d22316e71a4ae0b2cd8246b83771931e2bd778e4/dpgen/remote/decide_machine.py#L21 . The whole lines from 47 to 322 in decide_machine.py has been deprecated. No one still uses the function to automatically decide which machine to be used now.  The original codes are a mess.   So I commented all of them. We should put this function in `dpdispatcher` in a better way later in an elegant.

examples/CH4-refact-dpdispatcher/machine-ali-ehpc.json
AnguseZhang: No. Should be user_backward_files. In machine.json, we have left only three keys, `train`, `model_devi` and `fp`, each is a list of dict.  Each dict is a complete machine parameter, the design of list serves to keep the possibility for multiple machines.  Please refer to https://github.com/deepmodeling/dpgen/blob/devel/examples/CH4-refact-dpdispatcher/machine-dpcloudserver.json .  Old examples have been deprecated for a long time.

While in the main codes dpdispatcher still read the keys "model_devi_resources", "model_devi_machines",  we do a conversion in `convert_mdata` (in the past we use decide_fp_machine, decide_train_machine and decide_model_devi_machine ). In refactoring codes, the variables such as "model_devi_resources" should also be deprecated. Otherwise it's difficult to modularize the codes and generalize.

examples/init/INCAR_methane.rlx
AnguseZhang: This INCAR is the original version of our methane tutorials for DP-GEN.  I'm reorganizing the tutorials, so I put the original files.  In methane I chose a very relaxed structure as initial structure, so "IBRION=-1"  is enough, there's no need to move ions. But I admit that "IBRION=2" and "NSW=50" may be general

dpgen/generator/run.py
amcadmus: OK I see the point

dpgen/remote/decide_machine.py
amcadmus: I would suggest using '#' in front of all lines of the function, otherwise I cannot check which function is commented in an environment that does not have an auto highlighting like github

examples/init/INCAR_methane.rlx
amcadmus: The INCAR file has an extension `.rlx`, which indicates that it is for relaxation, so it is better to set IBRION=2. Otherwise if the INCAR is not for relaxation, we should remove `.rlx`

njzjz (approved):

examples/init/INCAR_methane.rlx
AnguseZhang: Fixed

AnguseZhang:
I also add support this function in `init_bulk`.  
Same support for `init_surf` and `auto_test` will be added in the future.

I also fixed bug in #487

examples/init/INCAR_methane.md
```diff
@@ -1,21 +1,33 @@
...
-NELM=120
-NELMIN=4
+
+NELM=200
+ISTART=0
+ICHARG=2
 ISIF=2
-ISMEAR=0
-SIGMA=0.20000
+ISMEAR=1
```
amcadmus, line 14: Why changing ISMEAR to 1?

tests/tools/context.py
```diff
@@ -2,9 +2,11 @@
 
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..')))
 from dpgen.tools.run_report import *
-
+machine_file = 'machine_fp_single.json'
```
amcadmus, line 5: Different tests may use different machine_file. I would not recommend define the it in the context.py, which serves to setup the environment and define  functions for global usage.

tests/tools/context.py
```diff
@@ -2,9 +2,11 @@
...
 def my_file_cmp(test, f0, f1):
     with open(f0) as fp0 :
         with open(f1) as fp1:
             test.assertTrue(fp0.read() == fp1.read())
 
+def setUpModule():
+    os.chdir(os.path.abspath(os.path.dirname(__file__)))
```
amcadmus, line 12: The name of the function is not corresponding to what it is doing

amcadmus (changes requested):

tests/tools/context.py
AnguseZhang: This is copied from https://github.com/deepmodeling/dpgen/blob/da494650314b3eebcf31ff20a7c195b309cb1a6a/tests/generator/context.py#L32

tests/tools/context.py
AnguseZhang: Fixed

examples/init/INCAR_methane.md
AnguseZhang: Thanks, you're right. should be IMSEAR = 0 for insulators like methane, fixed

tests/tools/context.py
AnguseZhang: Do you think we should modify them together.

amcadmus: Agree with you. I will fix other `setUpModule` in another PR.

amcadmus (approved):

tests/tools/context.py
amcadmus: The name is reserved for UTs and helps to change dir to the file dir. So we should not change it. Please see 
https://docs.python.org/3/library/unittest.html#setupmodule-and-teardownmodule

PR #486, opened 2021-08-04
exclude no converged task for cp2k fp

robinzyb:
Not converged cp2k fp tasks will not be included in data set


dpgen/generator/run.py
```diff
@@ -460,7 +460,7 @@ def run_train (iter_index,
     except KeyError:
         mdata = set_version(mdata)
 
-    
+
```
amcadmus, line 463: Please do not introduce such space removing revision. Thanks!

dpgen/generator/run.py
```diff
@@ -2503,20 +2503,29 @@ def post_fp_cp2k (iter_index,
...
-            sys = dpdata.LabeledSystem(oo, fmt = 'cp2k/output')
-            if len(sys) > 0:
-                sys.check_type_map(type_map = jdata['type_map'])
-            if idx == 0:
-                all_sys = sys
+        tcount += len(sys_output)
+        all_sys = None
+        for oo in sys_output :
+            _sys = dpdata.LabeledSystem(oo, fmt = 'cp2k/output')
+            if len(_sys) == 1:
```
njzjz, line 2515: dpdata should support appending empty system, so you don't need to make a if-statement...

AnguseZhang: I think this is used for counting failed ratio?

njzjz: I think it's more clear to use `len(sys_output)-len(all_sys)` instead of the current way.

dpgen/generator/run.py
robinzyb: My vim editor automatically removes the extra spaces, any suggestion in this scenario?

dpgen/generator/run.py
robinzyb: > I think it's more clear to use `len(sys_output)-len(all_sys)` instead of the current way.

I adapt this way

njzjz (approved):
Approve, but changes of removing space should be reverted.

robinzyb:
> Approve, but changes of removing space should be reverted.

how to effectively revert?  My vim editor automatically removes the extra spaces

njzjz:
I will recommend VS Code instead of vim...

dpgen/generator/run.py
robinzyb: all removed

PR #488, opened 2021-08-05
add conda publish github action

felix5572:
add conda publish github action.
it will automatically release dpgen to conda deepmodeling channel  (https://anaconda.org/deepmodeling/repo) when a new tag is added


conda/meta.yaml
```diff
@@ -0,0 +1,53 @@
+{% set name = "dpgen" %}
+{% set version = environ.get('GIT_DESCRIBE_TAG').lstrip('v') %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/deepmodeling/dpgen
+  # git_rev: {{ version }}
```
njzjz, line 10: Why do you comment this line?

.github/workflows/publish_conda.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: publish-to-conda
+      uses: felix5572/conda-publish-action@v1.9
+      with:
+        subdir: 'conda'
+        anacondatoken: ${{ secrets.ANACONDA_TOKEN }}
```
njzjz, line 17: Have you added the token?

conda/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+build:
+  number: 0
+  noarch: python
+  script: python -m pip install --no-deps --ignore-installed .
+
+requirements:
+  build:
+    - git
+  host:
+    - git
```
njzjz, line 21:
```suggestion
```

felix5572: fixed

.github/workflows/publish_conda.yml
felix5572: Haven't now, I don't have the permission to add repo secret.  We could add the secret later?

conda/meta.yaml
felix5572: typo.. uncomment now

conda/construct.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - defaults
+  - conda-forge
+  - deepmodeling
+
+specs:
+  - python 3.8
+  - pip
+  - dpgen {{ version }} 
+  - dpdata
+  - pymatgen
```
njzjz, line 16: only dpgen is enough

AnguseZhang (approved):
I have tested the patch on an independent machine, and things work well.  I have no other comments on the workflow besides Jinzhe's suggestions.

conda/meta.yaml
```diff
@@ -0,0 +1,53 @@
+{% set name = "dpgen" %}
+{% set version = os.popen('git describe --tags --abbrev=0').read().strip('\n').lstrip('v').replace('-', '_') %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/deepmodeling/dpgen
+  git_rev: v{{ version }}
```
njzjz, line 10: @felix5572 You can consider to use
```suggestion
  path: ..
```
instead. It will fetch source codes from the local file

njzjz (approved):

njzjz (approved):
Approve, but note that the secret should be added.

PR #493, opened 2021-08-09
Update gromacs model devi engine

Ericwang6:
1. Update gromacs model devi engine, supporting multi lambdas, multi temperatures of simulation.
2. Update gaussian fp engine, supporting different charges for different systems.
3. Bug fix of `UnboundLocalError` when parsing gaussian log


dpgen/generator/run.py
```diff
@@ -2483,18 +2534,22 @@ def post_fp_gaussian (iter_index,
...
         for idx,oo in enumerate(sys_output) :
-            sys = dpdata.LabeledSystem(oo, fmt = 'gaussian/log')
-            if len(sys) > 0:
-                sys.check_type_map(type_map = jdata['type_map'])
-            if jdata.get('use_atom_pref', False):
-                sys.data['atom_pref'] = np.load(os.path.join(os.path.dirname(oo), "atom_pref.npy"))
-            if idx == 0:
-                if jdata.get('use_clusters', False):
-                    all_sys = dpdata.MultiSystems(sys, type_map = jdata['type_map'])
+            # TODO : UnboundLocalError sometimes occurs and I cannot figure it out.
```
njzjz, line 2537: I don't think UnboundLocalError is an expected error that should be caught. If it happened, it is a bug that needs to be fixed.

dpgen/generator/run.py
```diff
@@ -183,9 +183,12 @@ def dump_to_poscar(dump, poscar, type_map, fmt = "lammps/dump") :
     sys = dpdata.System(dump, fmt = fmt, type_map = type_map)
     sys.to_vasp_poscar(poscar)
 
-def dump_to_deepmd_raw(dump, deepmd_raw, type_map, fmt='gromacs/gro'):
+def dump_to_deepmd_raw(dump, deepmd_raw, type_map, fmt='gromacs/gro', charge=None):
     system = dpdata.System(dump, fmt = fmt, type_map = type_map)
     system.to_deepmd_raw(deepmd_raw)
+    if charge is not None:
+        with open(os.path.join(deepmd_raw, "charge"), 'w+') as f:
```
njzjz, line 190: Why using `w+` instead of `w`?

dpgen/generator/run.py
```diff
@@ -1086,6 +1101,16 @@ def _make_model_devi_native_gromacs(iter_index, jdata, mdata, conf_systems):
     else:
         model_devi_dt = jdata['model_devi_dt']
     nsteps = cur_job.get("nsteps", None)
+    lambdas = cur_job.get("lambdas", [])
+    temps = cur_job.get("temps", [])
```
njzjz, line 1105: You can put the default value here... like
```suggestion
    lambdas = cur_job.get("lambdas", [1.])
    temps = cur_job.get("temps", [298.])
```
and remove lines 1106-1107, 1112-1113

njzjz (changes requested):

dpgen/generator/run.py
Ericwang6: Sure. I met this problem this week but I don't have time to fix it :(
I will remove this part and try to fix it afterwards.

dpgen/generator/run.py
Ericwang6: It's my own habits. It seems there are not so many differences between them ?

dpgen/generator/run.py
Ericwang6: Got it.

dpgen/generator/run.py
njzjz: `w+` will create both reading and writing buffers instead of only writing buffer.

njzjz (approved):

dpgen/generator/run.py
```diff
@@ -205,7 +208,14 @@ def make_train (iter_index,
...
     training_reuse_iter = jdata.get('training_reuse_iter')
     training_reuse_old_ratio = jdata.get('training_reuse_old_ratio', None)
-    training_reuse_stop_batch = jdata.get('training_reuse_stop_batch', 400000)
+
+    if 'training_reuse_stop_batch' in jdata.keys():
+        training_reuse_stop_batch = jdata['training_reuse_stop_batch']
+    elif 'training_reuse_numb_steps' in jdata.keys():
+        training_reuse_stop_batch = jdata['training_reuse_numb_steps']
+    else:
+        training_reuse_stop_batch = 40000
```
AnguseZhang, line 217: This defaulted value is expected to be 400000?

dpgen/generator/run.py
```diff
@@ -1086,6 +1101,13 @@ def _make_model_devi_native_gromacs(iter_index, jdata, mdata, conf_systems):
     else:
         model_devi_dt = jdata['model_devi_dt']
     nsteps = cur_job.get("nsteps", None)
+    lambdas = cur_job.get("lambdas", [1.0])
+    temps = cur_job.get("temps", [298.0])
+
+    for ll in lambdas:
+        if ll > 1:
+            raise RuntimeError("lambda is larger than 1.0")
```
AnguseZhang, line 1109: assert lambda in [0,1] ?

dpgen/generator/run.py
```diff
@@ -205,7 +208,14 @@ def make_train (iter_index,
     training_init_model = jdata.get('training_init_model', False)
     training_reuse_iter = jdata.get('training_reuse_iter')
     training_reuse_old_ratio = jdata.get('training_reuse_old_ratio', None)
-    training_reuse_stop_batch = jdata.get('training_reuse_stop_batch', 400000)
+
+    if 'training_reuse_stop_batch' in jdata.keys():
+        training_reuse_stop_batch = jdata['training_reuse_stop_batch']
+    elif 'training_reuse_numb_steps' in jdata.keys():
+        training_reuse_stop_batch = jdata['training_reuse_numb_steps']
```
AnguseZhang, line 215: What's is the reason for adding such alias?

AnguseZhang (changes requested):
Examples including sys_charges should be updated

dpgen/generator/run.py
Ericwang6: Yes

dpgen/generator/run.py
Ericwang6: OK

dpgen/generator/run.py
Ericwang6: Because in deepmd-kit 2.0, "stop_batch" is changed to "numb_steps".

AnguseZhang (approved):

PR #495, opened 2021-08-11
Adaptive trust levels

amcadmus:
Adaptively determines lower trust levels. In the input file one set the following keys to enable the adaptive trust levels. 
```json
    "model_devi_adapt_trust_lo":true,
    "model_devi_numb_candi_f":	50,
    "model_devi_numb_candi_v":	50,
    "model_devi_perc_candi_f":	5,
    "model_devi_perc_candi_v":	5,
```
Note that with `model_devi_adapt_trust_lo == true` the `model_devi_f_trust_lo` and `model_devi_v_trust_lo` will be ignored.

For each system, the `dpgen` makes two sets:
1. From the frames with force model deviation lower than `model_devi_f_trust_hi`, select `max(model_devi_numb_candi_f, model_devi_perc_candi_f*n_frames)` frames with largest force model deviation. 
2. From the frames with virial model deviation lower than `model_devi_v_trust_hi`, select `max(model_devi_numb_candi_v, model_devi_perc_candi_v*n_frames)` frames with largest virial model deviation. 

and then unite the two sets as the candidate set. The lowest model deviations of the candidate set are treated as lower trust levels. 

The adapted lower trust levels are printed so the convergence of `dpgen` can be checked as the change of lower levels in iterations.


njzjz (approved):

dpgen/generator/run.py
```diff
@@ -1321,11 +1322,169 @@ def check_bad_box(conf_name,
...
+                        for jj in idx_rest_accurate:
+                            fp_rest_accurate.append([tt, cc, jj])
+                    counter['accurate'] += len(idx_rest_accurate)
+                    idx_rest_failed = np.where(all_conf[ii][7:] >= f_trust_hi)[0]
+                    if detailed_report_make_fp:
+                        for jj in idx_rest_failed:
+                            fp_rest_failed.append([tt, cc, jj])
+                    counter['failed'] += len(idx_rest_failed)
+    
+    return fp_rest_accurate, fp_candidate, fp_rest_failed, counter
```
njzjz, line 1384: `fp_rest_accurate` and `fp_rest_failed` are unbounded when `detailed_report_make_fp` is `False`

wanghan-iapcm: expected to be `[]`s ?

njzjz: Any value is fine as they are not used anywhere else.

PR #496, opened 2021-08-11
Support relative force model deviation by normalizing the RMS force magnitude 

amcadmus:


njzjz:
This should only work for LAMMPS. Is it possible to merge some codes to dpdata?

njzjz (approved):

amcadmus:
> This should only work for LAMMPS. Is it possible to merge some codes to dpdata?

The key difficulty is to construct a LabeledSystem from lammps dump file, because no energy and virial information is recorded.

PR #562, opened 2021-10-06
Change mv to cp when downloading softlinks.

AnguseZhang:
If the file to be downloaded is a softlink, `cp` should be performed instead of `mv`. Otherwise, `lfile` is still a file linked to some original file, and when this file's removed, `lfile` will be invalid.


AnguseZhang:
@felix5572 This should also be applied to `dpdispatcher`.

njzjz (approved):

PR #565, opened 2021-10-08
update abacus-scf

maki49:


dpgen/generator/lib/abacus_scf.py
```diff
@@ -13,15 +13,29 @@ def make_abacus_pw_scf_kpt(fp_params):
         ret += str(k_points[i]) + " "
     return ret
 
-def make_abacus_pw_scf_input(fp_params):
+def make_abacus_scf_input(fp_params):
```
amcadmus, line 16: What is the reason of removing "pw"?  does the current implementation support both pw and atomic basis?

PR #566, opened 2021-10-08
add CH4 lebesgue example

KZHIWEI:


Liu-RX (commented):
I think you should set up three new folders in `examples/init`, `examples/machine`, `examples/run` and put the `init.json`, `machine.json`, `run_param.json` in them respectively. In this way you can keep in line with the existing example files.

PR #568, opened 2021-10-11
fix dispatcher bugs in autotest run

shazj99:
After commit 66492d10, autotest  has been broken down since it only submit one task to run actually. Here we revert it back to previous behavior.

Change-Id: I55a67580bd95d169ce596ed6f0ed2e36485ced9e


AnguseZhang (approved):
Thanks for your contributions!

amcadmus (approved):

PR #569, opened 2021-10-11
autotest: fix referenced before assignment error

shazj99:
Change-Id: I7f2fcc27353d9aa3cc95a1ec07010c9ec7a7cba2


AnguseZhang (commented):
Does this change really work? If https://github.com/deepmodeling/dpgen/blob/84731c704fab1b8ee88ebe858448597ec7b962f6/dpgen/auto_test/common_prop.py#L165 is not defined, function will return  in https://github.com/deepmodeling/dpgen/blob/84731c704fab1b8ee88ebe858448597ec7b962f6/dpgen/auto_test/common_prop.py#L163.  If `ret` is `None` at last,  `ret and ret.successful() ` seems also to fail.

PR #580, opened 2021-10-26
add fp_async_check_ratio param to enable async check for fp jobs

shazj99:
While running dpgen jobs, we found the duration time of FP phrase is a large part in each iteration. This is because the DFT computing for some candidates are really hard and time consuming, and we need to wait all those long tail to be finished before going to the next iteration. We found that the proportion of those candidates is very small, which may less than 1%.

So we add a new param `fp_async_check_ratio` to optimize the execution: if 99% of candidates have finished the FP step for 600 seconds, then we put the job of checking remains to an async process. Obviously, the data will be skipped in the coming `train` process, but we will add it back in next iteration. It may happen that those candidates are lost for some reason such as unexpected error or no retry in async process, but we think it is acceptable as the the ratio is very small.

Following is our test number, which significantly saving time: 
image

Usage(add in param.json):  
`
"fp_async_check_ratio": 0.99,
`

Change-Id: I448708b6b29635d172f40bf61b7c0a6397832b5b

### DIFF
--- a/dpgen/dispatcher/Dispatcher.py
+++ b/dpgen/dispatcher/Dispatcher.py
@@ -68,6 +68,16 @@ class Dispatcher(object):
             raise RuntimeError('unknown batch ' + batch_type)
         self.jrname = job_record
 
+    def do_async_check(self,
+                       job_handler,
+                       mark_failure):
+        finished_job, all_job = self.all_finished(job_handler, mark_failure)
+        dlog.debug("%d out of %d jobs will be checked in async mode" % (finished_job, all_job))
+        while finished_job != all_job:
+            time.sleep(60)
+            finished_job, all_job = self.all_finished(job_handler, mark_failure)
+        dlog.debug("all %d jobs have been checked finish in async mode" % finished_job)
+
     def run_jobs(self,
                  resources,
                  command,
@@ -80,7 +90,8 @@ class Dispatcher(object):
                  forward_task_deference = True,
                  mark_failure = False,
                  outlog = 'log',
-                 errlog = 'err') :
+                 errlog = 'err',
+                 async_check = 1) :
         job_handler = self.submit_jobs(resources,
                                        command,
                                        work_path,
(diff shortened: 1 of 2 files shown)


dpgen/dispatcher/Dispatcher.py
```diff
@@ -92,11 +103,21 @@ def run_jobs(self,
...
-        while not self.all_finished(job_handler, mark_failure) :
-            time.sleep(60)
-        # delete path map file when job finish
-        # _pmap.delete()
+        finished_job, all_job = self.all_finished(job_handler, mark_failure)
+        cnt = 0
+        while finished_job != all_job:
+            finished_ratio = float(finished_job) / float(all_job)
+            if finished_ratio >= async_check:
+                if cnt >= 10:
```
AnguseZhang, line 111: The reason for introducing `cnt` and `multiprocessing` here should be explained within comments.

AnguseZhang: I'm trying to understand how the async mode is achieved.  So if the condition `finished_ratio >= async_check`  is satisfied,  you will check if there is more tasks finished and try to collect them for ten times?  Is `multiprocessing` necessary?  Sometimes `multiprocessing` on my Mac is not stable, so I would avoid using it unless necessary.

dpgen/dispatcher/Dispatcher.py
```diff
@@ -233,7 +254,7 @@ def all_finished(self,
                     job_record.record_finish(cur_hash)
                     job_record.dump()
         job_record.dump()
-        return job_record.check_all_finished()
+        return job_record.check_all_finished(), len(job_record.record)
```
AnguseZhang, line 257: I think modifying return values in `all_finished` is a breaking change for check_dispatcher_status in DispatcherList

AnguseZhang (changes requested):
Thanks for your contributions. Although this might improve the efficiency of `fp` process in `DP-GEN`, I have a lot of concerns for the instability it introduces. 

First, it changes the logic for each iterations' data collection. In older versions, one set of `fp` data won't be changed any more if this `fp` process finishes. But now, any set of data may accumulate, and unit test is lacked for such a complicated logic workflow.

Second, I argee with @njzjz . `Dispatcher` in `dpgen` itself will be deprecated soon. We recommend users not to use the older version of `Dispatcher`. If we want to add some new features, we should add it into `dpdispatcher`. 

So I wonder if there is a better way to implement the feature you request? For example, can we combine the async mode in `mark_failure` parameters in `dpdispatcher` ? 

Yuzhi Zhang

njzjz:
Old dispatcher has been removed from DP-GEN. Please contribute to the new dpdispatcher instead.

PR #583, opened 2021-11-02
model_devi_clean_traj support int type

KZHIWEI:
if model_devi_clean_traj is integer, then the most recent n iterations will be retained, others will be removed.


AnguseZhang (changes requested):
Thanks for contributions!
Please update README correspondingly.

KZHIWEI:
Thank you for the review. I have update description of model_devi_clean_traj in the README.

AnguseZhang (approved):

PR #590, opened 2021-11-13
Added ABACUS pw interface for dpgen init_bulk

Liu-RX:


Liu-RX:
It seems that dpdata installed in the test does not support abacus/md yet, leading to the failure of the test.

wanghan-iapcm:
Could you please merge the recent devel branch? Thanks!

Liu-RX:
> Could you please merge the recent devel branch? Thanks!

Thank you for reminding me of this! I've already merged the pr with devel branch.

README.md
```diff
@@ -188,7 +188,12 @@ If you want to specify a structure as starting point for `init_bulk`, you may se
 ```json
 "from_poscar":	true,
 "from_poscar_path":	"....../C_mp-47_conventional.POSCAR",
-```
+```
```
wanghan-iapcm, line 191: Please notice that the format is not correct.

dpgen/data/tools/hcp.py
```diff
@@ -24,3 +24,12 @@ def poscar_unit (latt) :
...
 
+def STRU_unit(latt):
+
+    box = gen_box() * latt/np.sqrt(2)
+    stru_data = {}
+    stru_data['cells'] = box
+    stru_data['atom_numbs'] = [2]
+    stru_data['coords'] = np.matmul(np.array([[0.0, 0.0, 0.0], [1./3, 1./3, 1./2]]), box)
+    stru_data['types'] = [0, 0]
+    return stru_data
```
wanghan-iapcm, line 35: Why not using dpdata for converting from POSCAR to STRU? You do not need to write support for abacus for each crystal type.

README.md
Liu-RX: Sorry for that... I just fixed the bug.

dpgen/data/tools/hcp.py
Liu-RX: > 

😯  Sorry I did not realize it when I was writing the code. I will remove the STRU_unit codes and use the dpdata conversion code.

Liu-RX: I've removed the STRU_unit functions and replaced them with poscar_unit and dpdata conversion.

README.md
```diff
@@ -211,6 +215,10 @@ The bold notation of key (such as **Elements**) means that it's a necessary key.
 | **md_nstep** | Integer | 10 | Steps of AIMD in stage 3. If it's not equal to settings via `NSW` in `md_incar`, DP-GEN will follow `NSW`.
 | **coll_ndata** | Integer | 5000 | Maximal number of collected data.
 | type_map | List | [ "Mg", "Al"] | The indices of elements in deepmd formats will be set in this order.
+| init_fp_style | String | "ABACUS" or "VASP" | First-principle software. If this key is abscent, the default value will be "VASP".
+| relax_kpt | String | "....../KPT" | Path of `KPT` file for relaxation in stage 1. Only useful if `init_fp_style` is "ABACUS".
+| md_kpt | String | "....../KPT" | Path of `KPT` file for MD simulations in stage 3. Only useful if `init_fp_style` is "ABACUS".
```
wanghan-iapcm, line 220: It is not good idea to explicitly set the number of kpoint, because the k-mesh should be adjusted according to the size of the system. For example different `super_cell` settings needs different number of kpoints. 

A better and recommended way of setting kmesh is to set kspacing. you can check `auto_test.lib.vasp.make_kspacing_kpoints` to make a kmesh from a given kspacing.

dpgen/data/tools/diamond.py
```diff
@@ -25,3 +25,12 @@ def poscar_unit (latt) :
...
 
+def STRU_unit(latt):
+
+    box = gen_box() * latt
+    stru_data = {}
+    stru_data['cells'] = box
+    stru_data['atom_numbs'] = [2]
+    stru_data['coords'] = np.matmul(np.array([[0.125, 0.125, 0.125], [0.875, 0.875, 0.875]]), box)
+    stru_data['types'] = [0, 0]
+    return stru_data
```
wanghan-iapcm, line 36: This method should not be needed.

README.md
Liu-RX: Sorry for the delay. I did not receive the notifications orz. I will solve the two issues as suggested.

dpgen/data/tools/diamond.py
Liu-RX: I've removed it.

README.md
Liu-RX: I have added `relax_kspacing` and `md_kspacing` in `param.json` for dpgen to generate the `KPT` files. I also reserved the previous `relax_kpt` and `md_kpt` parameters. Both methods are available.

wanghan-iapcm (approved):

PR #593, opened 2021-11-20
change compress level to 6 from default 9 to accelerate compress

taipinghu:
In some case, FP calculations could read the wave function file or charge files (e.g., CHGCAR and WAVECAR in VASP ). If the size of those files is too big, the compress process could be very slow due to the default setting (compresslevel=9) in tarfile module of python. Change the compresslevel to 6 will accelate compress rate.

### DIFF
--- a/dpgen/dispatcher/SSHContext.py
+++ b/dpgen/dispatcher/SSHContext.py
@@ -280,7 +280,7 @@ class SSHContext (object):
         os.chdir(self.local_root)
         if os.path.isfile(of) :
             os.remove(of)
-        with tarfile.open(of, "w:gz", dereference = dereference) as tar:
+        with tarfile.open(of, "w:gz", dereference = dereference, compresslevel=6) as tar:
             for ii in files :
                 tar.add(ii)
         os.chdir(cwd)


wanghan-iapcm (approved):

wanghan-iapcm:
@felix5572 @njzjz  Is this change submitted to the dpdispatcher?

njzjz:
> @felix5572 @njzjz  Is this change submitted to the dpdispatcher? 

Not yet but it should.

wanghan-iapcm:
@njzjz Do you have any idea why the UTs failed?

njzjz:
> @njzjz Do you have any idea why the UTs failed?

Maybe a dependency is lacked in dpdata. I'll add it.

njzjz:
> > @njzjz Do you have any idea why the UTs failed?
> 
> Maybe a dependency is lacked in dpdata. I'll add it.

https://github.com/deepmodeling/dpdata/pull/227

wanghan-iapcm:
> > > @njzjz Do you have any idea why the UTs failed?
> > 
> > 
> > Maybe a dependency is lacked in dpdata. I'll add it.
> 
> [deepmodeling/dpdata#227](https://github.com/deepmodeling/dpdata/pull/227)

Why it was not triggered by the UTs of dpdata?

njzjz:
It is a dependency of something else? It was indeed installed in the UT workflow of dpdata.

wanghan-iapcm:
I see. strange enough...

PR #595, opened 2021-11-24
bump dpdata version to 0.2.4

wanghan-iapcm:


AnguseZhang (approved):

njzjz:
I can fix the error in `TestPostGaussian`, but where does the error in `TestPostCP2K` come from?

PR #596, opened 2021-11-24
Fix bug of fp_task_min

AnguseZhang:
Currently  fp_task_min doesn't work.
The reason is `task_min` is set to -1 and passed to  https://github.com/deepmodeling/dpgen/blame/v0.10.2/dpgen/generator/run.py#L1538 .


wanghan-iapcm (commented):
The idea of `fp_task_min` is not to skip fp calculations when the number of fp jobs is too small, but to skip the training in the next iteration if the number of new fp jobs is too small.  In this case, the fp tasks are labeled, and the labels are stored for future trainings.

lauthirteen:
Dear Teacher Wang, 
        If the number of fp jobs is too small in this fp calculation, and it is not added in all subsequent iterations?  Or, what does the "future trainings" mean here ?

lauthirteen:
> The idea of `fp_task_min` is not to skip fp calculations when the number of fp jobs is too small, but to skip the training in the next iteration if the number of new fp jobs is too small. In this case, the fp tasks are labeled, and the labels are stored for future trainings.

Dear Teacher Wang,
        If the number of fp jobs is too small in this fp calculation, and it is not added in all subsequent iterations? Or, what does the "future trainings" mean here ?

PR #603, opened 2021-11-27
debug iterdict function for generate cp2k input

ChiahsinChu:
The original `iterdict` function used for generating CP2K input finds the place for inserting a new string by `list.index(object)`. This method fails when there are two identical keywords/section names in the input file (but at different levels). For example, there are many `PRINT` sections available in CP2K, which cannot be printed out correctly with the original `iterdict` function. 

In the modified version, I replace the Arg flag with loop_idx for recording the exact positions where the new string should be inserted (from the end of the list).
A minor change in the `make_cp2k_input` function is made correspondingly.


robinzyb:
Sorry I missed this PR from last year. i can't see why the pr fail the unittest. That is weird. @ChiahsinChu

PR #606, opened 2021-11-29
bug_fix | to support multi-element cases

Vibsteamer:
currently, only count the #atoms of the first element when evaluating E-per-atom;


Vibsteamer:
wrong auto-generate-title for the second commit and can't fix it.
re-patch in another pr

PR #608, opened 2021-12-01
bug_fix | to support multi-element cases

Vibsteamer:
currently, only count the #atoms of the first element when evaluating E-per-atom;


Vibsteamer:
re-patch, the same as pr #606 (closed due to the wrong title of the add-on commit within)

kevinwenminion (approved):

PR #610, opened 2021-12-07
model_devi | sort models in default&support user-defined order

Vibsteamer:
current behavior:
1）the list of models written in input.lammps for model_devi is created by "glob" without "sort" --> may give different (random) orders in different machine environments  --> would randomly alter the model being used to sample configurations (always use the first model written in imput.lammps)
---> could lead to unexpected performance, such as, when using together with "model_devi_activation_func" that allows four models to be nonequivalent.
2）when preparing input.lammps from iuser-provided template, the line begin with "pair_style deepmd" will be overwritten by dpgen, thus overwrites the user defined order of models ---> could lead to unexpected performance, such as, when using together with "model_devi_activation_func" that allows four models to be nonequivalent and users indeed expected a specific order of models.

changes:
1）sorted the list of models in default, thus always use graph.000.pb to sample configurations;
2）check weather user writes the full line of begin with "pair_style deepmd" (by checking the number of non-space words in this line),  if yes, leave it be; if not, overwrites with the default settings (use graph.000.pb to sample); besides, the original error trigger is retained if key words "pair_style deepmd" are not provided in the template.

### DIFF
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -947,7 +947,7 @@ def _make_model_devi_revmat(iter_index, jdata, mdata, conf_systems):
                 # revise input of lammps
                 with open('input.lammps') as fp:
                     lmp_lines = fp.readlines()
-                # only revise the line "pair deepmd" if the user has not written the full line (checked by then length of the line)
+                # only revise the line "pair_style deepmd" if the user has not written the full line (checked by then length of the line)
                 template_has_pair_deepmd=1
                 for line_idx,line_context in enumerate(lmp_lines):
                     if (line_context[0] != "#") and ("pair_style" in line_context) and ("deepmd" in line_context):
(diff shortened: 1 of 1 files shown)


AnguseZhang (approved):

AnguseZhang:
Could please upload or update the examples for revmat-mode? Thanks!

Vibsteamer:
> Could you please upload or update the examples for revmat-mode? Thanks!

will be uploaded afterwards

wanghan-iapcm:
There can be any number of space between "pair_style" and "deepmd", so the length check may fail.

Vibsteamer:
> There can be any number of space between "pair_style" and "deepmd", so the length check may fail.

Sorry for the misleading description of the pr.

It actually checks the 【numbers of non-space words in the line】 by 【.split()】, 
thought this works for the any number of space in the line

```
if len(lmp_lines[template_pair_deepmd_idx].split()) !=  (len(models) + len(["pair_style","deepmd","10", "model_devi.out"])):
```

PR #615, opened 2021-12-12
Add gromacswrapper to conda recipe

AnguseZhang:


conda/gromacswrapper/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  host:
+    - pip
+    - python
+  run:
+    - matplotlib-base
+    - numkit
```
njzjz, line 24: This package is not available yet.
I think you can submit to https://github.com/conda-forge/staged-recipes instead? We do not need to maintain every package by ourselves.

AnguseZhang: > 

I agree with you.  
We can submit the recipes to `conda-forge`. But I found some PRs have been in progressing for a long time before it can be merged into `conda-forge`, so I also submitted numkit recipe, which can solve the problem currently.  @felix5572 Maybe would you like to join this?

njzjz: However, I don't think it needs to be summited to DP-GEN's repository.

njzjz (changes requested):
A problem here is that we need to build the dependencies and upload to deepmodeling's channel. However, adding them here won't build and upload automatically, so I do not see the meaning of this PR.

conda/gromacswrapper/meta.yaml
AnguseZhang: @njzjz We decide to publish `0.10.4` version tomorrow, and I haven't submitted the  recipes to `conda-forge` yet.  But I  still hope to merge this PR first to let offline package be available for the community. What's your opinion ?

njzjz: You also need to publish packages to our channel

njzjz:
These two recipes are added to
- https://github.com/deepmd-kit-recipes/numkit-feedstock
- https://github.com/deepmd-kit-recipes/gromacswrapper-feedstock

This PR could be closed.

AnguseZhang:
> These two recipes are added to
> 
> * https://github.com/deepmd-kit-recipes/numkit-feedstock
> * https://github.com/deepmd-kit-recipes/gromacswrapper-feedstock
> 
> This PR could be closed.

netCDF4's dependency should be added. https://github.com/deepmodeling/dpgen/pull/792/files

PR #643, opened 2022-02-17
Add dpgen analysis module 

taipinghu:
In order to achieve more convenient analysis in dpgen, add analysis module in dpgen. dpgen analysis is a interactive analysis tools. You can see the screen to understand each function.

### DIFF
--- a/dpgen/analysis/main_func_1.py
+++ b/dpgen/analysis/main_func_1.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+
+from dpdata import LabeledSystem
+
+from dpgen.analysis.analysis_utils import select_logs, process_bar
+
+###########################################################
+def data_collections():
+    while True:
+        print("\n")
+        print("========================================================")
+        print("                    Main Menu 1                        ")
+        print("                                                       ")
+        print(" ( 11)  Collect all data in current dir from OUTCAR file")
+        print("\n Tips: Input q or -10 to exit program                ")
+        print("========================================================")
+        
+        jsel = input("\n Please input the menu index: ")
+        if jsel == str(-10):
+            break
+        elif jsel == str(11):
+            root_dir = input("\n Please input the root dir name: ")
+            str_in   = input("\n Collect (1) all frames (e.g., AIMD task) or (2) final frame (e.g., Optimization task)? ")
+            out_dir  = input("\n Please input the out dir name: ")
+            total_frames = -1
+            if str_in == str(1):
+                total_frames = 0
+            collect_all_data_in_current_dir_from_all_OUTCAR_file(root_dir=root_dir, total_frames=total_frames, out_dir=out_dir)
+###########################################################
+
+
+###########################################################
+def collect_all_data_in_current_dir_from_all_OUTCAR_file(root_dir: str, total_frames: int, out_dir: str) -> LabeledSystem:
+    all_outcar = select_logs(root_dir, "OUTCAR")
+    print(" Totally find %d OUTCAR in current dir."%len(all_outcar))        
+
+    final_sys = None
+    counter = 0
+    for outcar in all_outcar:
+        counter += 1
+        labeled_system = LabeledSystem(outcar, fmt="vasp/outcar")
+        if total_frames == -1:
+            frames = [-1, ]
+        else:
+            frames = range(labeled_system.get_nframes())
+        for iframe in frames:
+            if final_sys is None:
+                final_sys = labeled_system[iframe]
+            else:
+                final_sys.append(labeled_system[iframe])
+        process_bar(counter, len(all_outcar), " Read %s"%outcar)
+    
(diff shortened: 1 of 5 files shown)


dpgen/analysis/main_func_1.py
```diff
@@ -0,0 +1,55 @@
...
+        print("\n Tips: Input q or -10 to exit program                ")
+        print("========================================================")
+        
+        jsel = input("\n Please input the menu index: ")
+        if jsel == str(-10):
+            break
+        elif jsel == str(11):
+            root_dir = input("\n Please input the root dir name: ")
+            str_in   = input("\n Collect (1) all frames (e.g., AIMD task) or (2) final frame (e.g., Optimization task)? ")
+            out_dir  = input("\n Please input the out dir name: ")
```
wanghan-iapcm, line 24: It is highly recommended to use `argparse` not `input` to process input arguments for dpgen commands.

wanghan-iapcm (commented):
This command seems to "Collect all data in current dir from OUTCAR file". I would recommend not using the command name `analysis`, because it is not self-explanatory. 

What is the difference between `dpgen collect` and the `dpgen analysis`? Both command collect labeled data from dpgen iterations.

taipinghu:
> This command seems to "Collect all data in current dir from OUTCAR file". I would recommend not using the command name `analysis`, because it is not self-explanatory.
> 
> What is the difference between `dpgen collect` and the `dpgen analysis`? Both command collect labeled data from dpgen iterations.

Collect data is one of function of dpgen analysis. Our team will add more functions in the dpgen analysis. Similar functions can be found in DPAnalysis (https://github.com/taipinghu/DPAnalysis). According to the Yuzhi's suggestions, DPAnalysis can be integrated into dpgen latter.

dpgen/analysis/main_func_1.py
taipinghu: > It is highly recommended to use `argparse` not `input` to process input arguments for dpgen commands.

we think it is more conveniet to analysis for users. Users donot need to know the specific argment. Instead, they can use dpgen analysis according to the output the screen.

AnguseZhang: This will 

> > It is highly recommended to use `argparse` not `input` to process input arguments for dpgen commands.
> 
> we think it is more conveniet to analysis for users. Users donot need to know the specific argment. Instead, they can use dpgen analysis according to the output in screen.

This will definitely reduce the flexibility for this function. For a user who's already familiar with the command and wishes to perform a series of tasks, such design will force him/her to manually type command in the interactive mode each time.

taipinghu: > > > It is highly recommended to use `argparse` not `input` to process input arguments for dpgen commands.
> > 
> > 
> > we think it is more conveniet to analysis for users. Users donot need to know the specific argment. Instead, they can use dpgen analysis according to the output in screen.
> 
> This will definitely reduce the flexibility for this function. For a user who's already familiar with the command and wishes to perform a series of tasks, such design will force him/her to manually type command in the interactive mode each time.

There's no need to worry about that. For batch analysis, users can prepare an input file containing the input in interactive mode and then perform analysis by redirection such as "dpgen analysis < input > output". For example, Multiwfn (http://sobereva.com/multiwfn/) is the most popular software for wavefunction analysis, , which uses interactive mode. For a analysis tool, it is impossible for users to remember every argument of each function.

njzjz:
dpdata has had a method to collect data from a directory. https://github.com/deepmodeling/dpdata/blob/81c656cc92cf37cca17249e506dc81bb6dfe5549/dpdata/system.py#L1130-L1136 It's unclear to me if it's meaningful to reimplement it.

taipinghu:
> dpdata has had a method to collect data from a directory. https://github.com/deepmodeling/dpdata/blob/81c656cc92cf37cca17249e506dc81bb6dfe5549/dpdata/system.py#L1130-L1136 It's unclear to me if it's meaningful to reimplement it.

Collect data is one of functions of dpgen analysis. 
This is first time to pr about dpgen analysis. we will pull more requests latter. Those functions can be seen in DPAnalysis.

dpgen/analysis/main_func_1.py
wanghan-iapcm: `dpgen analysis < input` works, but we prefer the argparse for three reasons: 
1. argparse allows arbitrary order of argument.
2. one can easy access the document for ALL arguments. 
3. argparse has type checking.
For the reasons, all other dpgen commands use argparse. Please keep the style of dpgen commands consistent.

njzjz:
> Collect data is one of functions of dpgen analysis. 
> 
> This is first time to pr about dpgen analysis. we will pull more requests latter. Those functions can be seen in DPAnalysis. 

This PR mainly focuses on converting a format to another format. It should be a cli of dpdata instead.

PR #696, opened 2022-04-10
add bcc interstitial calculation and change some default parameters in lmp

kevinwenminion:
1. Add bcc interstitial calculations for transition metals which have six types of interstitial positions.
2. In lmp relaxation, change the convergence to be controlled by the force criteria. This is more reasonable, especially for large systems.
3. Change the "pair_coeff" in lmp to be compatible with the latest lmp version.


wanghan-iapcm (commented):
Could you please add a UT for added the BCC interstitial?

kevinwenminion:
> Could you please add a UT for added the BCC interstitial?

Yes, I will add the unittest.

Issue #697, opened 2022-04-10
[BUG] _"dpgen autotest make relaxation.json" for deepmd


xuekuiwin:
**Summary**

When dpgen is run (dpgen autotest make relaxation.json), it forces a change to the in.lammps  parameter "pair_coeff   * *", and delete "**"
**DPGEN Version and Platform**
DPGEN :0.10.3; platform: pbs

**Job submission and computing cluster configuration**

**Expected Behavior**
add  "**"

**Actual Behavior**

you can see dir: "confs/std-hcp/relaxation/in.lammps" and "lammps_input/in.lammps"
**Steps to Reproduce**

relaxation.json:
{
    "structures":         ["confs/std-*"],
    "interaction": {
            "type":        "deepmd",
            "model":       "frozen_model.pb",
            "in_lammps":   "lammps_input/in.lammps",
            "type_map":   {"Ti": 0}
        },
    "relaxation": {
            "cal_type":    "relaxation",
            "cal_setting":{"relax_pos":       true,
                           "relax_shape":     true,
                           "relax_vol":       true,
                           "etol": 1e-12,
                           "ftol": 1e-6,
                           "maxiter": 5000,
                           "maximal": 500000}
        }
}

in.lammps:
clear
units             metal
dimension         3
boundary          p p p
atom_style        atomic
box               tilt large
read_data         conf.lmp
mass              1 47.88
neigh_modify      every 1 delay 0 check no
pair_style        deepmd frozen_model.pb
pair_coeff        * *
compute           mype all pe
thermo            100
thermo_style      custom step pe pxx pyy pzz pxy pxz pyz lx ly lz vol c_mype
dump              1 all custom 100 dump.relax id type xs ys zs fx fy fz
min_style         cg
fix               1 all box/relax iso 0.0
minimize          1.000000e-12 1.000000e-06 5000 500000
fix               1 all box/relax aniso 0.0
minimize          1.000000e-12 1.000000e-06 5000 500000
variable          N equal count(all)
variable          V equal vol
variable          E equal "c_mype"
variable          tmplx equal lx
variable          tmply equal ly
variable          Pxx equal pxx
variable          Pyy equal pyy
variable          Pzz equal pzz
variable          Pxy equal pxy
variable          Pxz equal pxz
variable          Pyz equal pyz
variable          Epa equal ${E}/${N}
variable          Vpa equal ${V}/${N}
variable          AA equal (${tmplx}*${tmply})
print "All done"
[...]

PR #698, opened 2022-04-10
Fix the usage of `pair_coeff` in `auto_test`

AnguseZhang:
See https://github.com/deepmodeling/dpgen/pull/567 and https://github.com/deepmodeling/dpgen/issues/697

### DIFF
--- a/dpgen/auto_test/lib/lammps.py
+++ b/dpgen/auto_test/lib/lammps.py
@@ -115,7 +115,7 @@ def inter_deepmd(param):
             ret += "%s out_freq 10 out_file model_devi.out\n" % model_list
         else:
             ret += models[0] + '\n'
-    ret += "pair_coeff\n"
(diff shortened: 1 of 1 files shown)


njzjz (changes requested):
Please also fix https://github.com/deepmodeling/dpgen/blob/9e3b958e07ebcc1858405daff1033b1a94d88f58/dpgen/auto_test/template/elastic/lmp/potential.mod#L7 and other related codes.

AnguseZhang:
> pair_coeff

Does a normal LAMMPS without DeePMD also require `pair_coeff * *`?

njzjz:
> > pair_coeff
> 
> 
> 
> Does a normal LAMMPS without DeePMD also require `pair_coeff * *`? @njzjz 

I think it't not support any more in a new version of LAMMPS.

wanghan-iapcm:
See also #696

PR #730, opened 2022-05-12
refactor simplify

njzjz:
1. Used `dp model-devi` to calculate model deviation, instead of local calculation. Supported by deepmodeling/deepmd-kit#1618, released in v2.1.1. So the version earlier than 2.1.1 is not supported any more.
2. Assumed all systems are MultiSystems.
3. Removed energy model deviation support


njzjz:
Paths are expanded in the `get_multi_system`, `make_train`, `run_train` method, whatever the parameter.

njzjz:
> Did we delete support for `use_clusters`? If true, we should also change the readme https://github.com/deepmodeling/dpgen/blob/05d1ac5354c888bff9cba73439fe129c7f25b10c/README.md .

It has been updated accordingly.

wanghan-iapcm (approved):

PR #741, opened 2022-05-30
update: add a new option of absolute volume in /dpgen/dpgen/auto_test/EOS.py

HuangJiameng:
update: add a new option of absolute volume in /dpgen/dpgen/auto_test/EOS.py

### DIFF
--- a/doc/toymodels/JiamengHuang_pr.md
+++ b/doc/toymodels/JiamengHuang_pr.md
@@ -6,4 +6,4 @@ in the "eos" part of property.json
 if it's not mentioned, "False" is set defaultly
 when you are using absolute volume, there will be a notation in the last line of output during "make" process, which is like
 
-(absolute volume)
+treat vol_start and vol_end as absolute volume
--- a/doc/toymodels/JiamengHuang_pr.md
+++ b/doc/toymodels/JiamengHuang_pr.md
@@ -1,0 +1,9 @@
+A new parameter "vol_abs" is added. If you want to use absolute volume to get EOS, you can add
+
+         "vol_abs":      true,
+
+in the "eos" part of property.json
+if it's not mentioned, "False" is set defaultly
+when you are using absolute volume, there will be a notation in the last line of output during "make" process, which is like
+
+(absolute volume)
--- a/dpgen/auto_test/EOS.py
+++ b/dpgen/auto_test/EOS.py
@@ -24,6 +24,8 @@ class EOS(Property):
                 self.vol_start = parameter['vol_start']
                 self.vol_end = parameter['vol_end']
                 self.vol_step = parameter['vol_step']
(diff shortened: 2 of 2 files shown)


AnguseZhang (changes requested):

dpgen/auto_test/EOS.py
```diff
@@ -117,6 +119,7 @@ def make_confs(self,
 
             else:
                 print('gen eos from ' + str(self.vol_start) + ' to ' + str(self.vol_end) + ' by every ' + str(self.vol_step))
+                if self.vol_abs : print('(absolute volume)')
```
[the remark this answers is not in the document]
HuangJiameng: For suggestion 1, since the user has already push the "vol_abs" button in property.json, maybe it's ok to be in "Description" part.
Suggestion 2 is implemented.

doc/toymodels/JiamengHuang_pr.md
```diff
@@ -0,0 +1,9 @@
+A new parameter "vol_abs" is added. If you want to use absolute volume to get EOS, you can add
```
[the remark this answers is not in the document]
AnguseZhang: I've added this to https://github.com/deepmodeling/dpgen/wiki/Property:-get-started-and-input-examples

dpgen/auto_test/EOS.py
```diff
@@ -24,6 +24,8 @@ def __init__(self,
                 self.vol_start = parameter['vol_start']
                 self.vol_end = parameter['vol_end']
                 self.vol_step = parameter['vol_step']
+                parameter['vol_abs'] = parameter.get('vol_abs', False)
```
njzjz, line 27: [`set_default`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) is better than `get`.

dpgen/auto_test/EOS.py
```diff
@@ -117,6 +119,10 @@ def make_confs(self,
 
             else:
                 print('gen eos from ' + str(self.vol_start) + ' to ' + str(self.vol_end) + ' by every ' + str(self.vol_step))
+                if self.vol_abs : 
+                    print('treat vol_start and vol_end as absolute volume')
```
njzjz, line 123: Use `dlog.info` instead.

njzjz (changes requested):

dpgen/auto_test/EOS.py
HuangJiameng: `get` is applied in other parts of dpgen. To keep consistency, I think it's ok.

njzjz: If you mean the other parts in this file, indeed I don't think they are ok. Here `set_default` is faster and more concise. I can approve this PR though.

njzjz (approved):

PR #748, opened 2022-06-08
supports non-list mdata

njzjz:
The mdata of a task is a list of a single dict. This "list" looks useless and not clear enough. So this commit supports using the dict without a list.

Note that old parameters are still supported, so no breaking changes are made. The "list" is just removed from all examples. Both list and non-list parameters are in the unit test.

Old parameter:
```
"train": [
  {
    "command": "dp"
  }
]
```
New parameter:
```
"train": {
    "command": "dp"
  }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1139,7 +1139,7 @@ an example of new dpgen's machine.json
 ```json
 {
   "api_version": "1.0",
-  "train": [
+  "train":
     {
       "command": "dp",
       "machine": {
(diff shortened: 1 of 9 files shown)


AnguseZhang (approved):
I just explain why it's designed to be a list.  Once I had three Slurm clusters, and tens of DP-GEN projects. So I want to make sure that DP-GEN submits its tasks to the most available (having most empty nodes or shortest waiting queue), and implement some detective functions in dispatcher. But it seems no one except me has this need, and so when we upgrade to `dpdispatcher` this function is deprecated.

PR #749, opened 2022-06-08
upgrade all tasks to dpdispatcher

njzjz:
This commit upgrades init_reaction and init_surf to use dpdispatcher

### DIFF
--- a/dpgen/dispatcher/Dispatcher.py
+++ b/dpgen/dispatcher/Dispatcher.py
@@ -1,5 +1,6 @@
 from distutils.version import LooseVersion
 import os,sys,time,random,json,glob
+import warnings
 from typing import List
 from dpdispatcher import Task, Submission, Resources, Machine
 from dpgen.dispatcher.LocalContext import LocalSession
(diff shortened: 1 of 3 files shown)


dpgen/dispatcher/Dispatcher.py
```diff
@@ -406,3 +407,81 @@ def mdata_arginfo() -> List[Argument]:
     return [
         command_arginfo, machine_arginfo, resources_arginfo,
     ]
+
+
+def make_submission_compat(
+        machine: dict,
+        resources: dict,
+        commands: List[str],
+        work_dir: str,
```
AnguseZhang, line 416: `work_dir` is the older version. In `dpdispatcher` version, this is `work_path`. Maybe we should let `work_path` be default?

dpgen/dispatcher/Dispatcher.py
```diff
@@ -406,3 +407,81 @@ def mdata_arginfo() -> List[Argument]:
...
+    forward_files : list[str]
+        forwarded files for each task
+    backward_files : list[str]
+        backwarded files for each task
+    outlog : str, default=log
+        path to log from stdout
+    errlog : str, default=err
+        path to log from stderr
+    api_version : str, default=0.9
+        API version. 1.0 is recommended
```
AnguseZhang, line 456: Do you think that we should upgrade the default to 1.0 ?

AnguseZhang, line 456: Do you think that we should upgrade the default to 1.0 ?

wanghan-iapcm (approved):

PR #750, opened 2022-06-08
Update README.md

pee8379:
'pert_box' dose not seems to work as percentage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ The following table gives explicit descriptions on keys in `PARAM`.
 
 The bold notation of key (such as **Elements**) means that it's a necessary key.
 
- Key  | Type          | Example                                                      | Discription                                                      |
+ Key  | Type          | Example                                                      | Description                                                      |
 | :---------------- | :--------------------- | :-------------------------------------- | :-------------------------------------------------------------|
 | **stages** | List of Integer | [1,2,3,4] | Stages for `init_bulk`
 | **Elements** | List of String | ["Mg"] | Atom types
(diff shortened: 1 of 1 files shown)


AnguseZhang (changes requested):
Thanks for your contribution!  Though, I do think `pert_box` serves as changing the percentage, instead of an absolute length for box length. (See  https://github.com/deepmodeling/dpgen/blob/0767dce7cad29367edb2e4a55fd0d8724dbda642/dpgen/data/tools/create_random_disturb.py#L93 ). How does it behave in your circumstance?

pee8379:
Right. I misunderstood, only saw  emat = emat + np.eye(3) and not 
cell = np.dot(cell0, gen_random_emat(etmax, diag))
part.

I originally had a interest this part, because some of my POSCARs generated by init_bulk has highly deformed cell. 
does 
"pert_box":	1
means 1% of changing in box? or should it be 0.01 to get 1% of change?

Thanks.

HuangJiameng:
> 

I see what you mean, maybe "ratio" is more appropriate than "percentage"
see 'pert_cmd = 'python3 ' + pert_cmd + ' -etmax %f -ofmt %s %s %d %f > /dev/null' %(pert_box, fp_style, poscar_name, pert_numb, pert_atom)' in [https://github.com/deepmodeling/dpgen/blob/0767dce7cad29367edb2e4a55fd0d8724dbda642/dpgen/data/gen.py](url)

njzjz:
I close this PR as the README has been fully refactored. If one wants to change the documentation, go to `argcheck.py`.

PR #765, opened 2022-07-01
fix ‘post_fp_cp2k’, add param rfailed

HuangJiameng:
1. ratio_failed was not applied in post_fp_cp2k.
2. When a cp2k task fails, it returns an empty "output" file. Since dpdispatcher only checks whether the output file exists or not, it marks the task finished and sends back the empty file. An empty or not converged output file leads to errors in the following step. Commenting out check_type_map is a temporary solution, which can run the next step by ignoring empty output files. To solve this problem fundamentally, dpdispatcher should probably be modified as well. 

- add ratio_failed param to cp2k in 02.fp step
- fix a bug in `post_fp_cp2k`

### DIFF
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -2995,7 +2995,10 @@ def post_fp_gaussian (iter_index,
 
 
 def post_fp_cp2k (iter_index,
-                      jdata):
+                      jdata,
+                      rfailed=None):
+                      
+    ratio_failed =  rfailed if rfailed else jdata.get('ratio_failed',0.05)
     model_devi_jobs = jdata['model_devi_jobs']
     assert (iter_index < len(model_devi_jobs))
 
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -2998,7 +2998,7 @@ def post_fp_cp2k (iter_index,
                       jdata,
                       rfailed=None):
                       
-    ratio_failed =  rfailed if rfailed else jdata.get('ratio_failed',0.05)
+    ratio_failed =  rfailed if rfailed else jdata.get('ratio_failed',0.10)
     model_devi_jobs = jdata['model_devi_jobs']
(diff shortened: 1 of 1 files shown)


dpgen/generator/run.py
```diff
@@ -3026,7 +3029,7 @@ def post_fp_cp2k (iter_index,
         all_sys = None
         for oo in sys_output :
             _sys = dpdata.LabeledSystem(oo, fmt = 'cp2k/output')
-            _sys.check_type_map(type_map = jdata['type_map'])
+            #_sys.check_type_map(type_map = jdata['type_map'])
```
wanghan-iapcm, line 3032: Could you explain why the check_type_map is commented out?

HuangJiameng: Thanks for your comment, I have edited my comment.

wanghan-iapcm (approved):

wanghan-iapcm:
> 2\. When a cp2k task fails, it returns an empty "output" file. Since dpdispatcher only checks whether the output file exists or not, it marks the task finished and sends back the empty file. An empty or not converged output file leads to errors in the following step. Commenting out check_type_map is a temporary solution, which can run the next step by ignoring empty output files. To solve this problem fundamentally, dpdispatcher should probably be modified as well.

@njzjz @felix5572 FYI

njzjz:
> > 2. When a cp2k task fails, it returns an empty "output" file. Since dpdispatcher only checks whether the output file exists or not, it marks the task finished and sends back the empty file. An empty or not converged output file leads to errors in the following step. Commenting out check_type_map is a temporary solution, which can run the next step by ignoring empty output files. To solve this problem fundamentally, dpdispatcher should probably be modified as well.
> 
> @njzjz @felix5572 FYI

This is not the expected behavior. dpdispatcher should check the return code, unless cp2k always returns 0.

PR #767, opened 2022-07-04
Fix simplify

HuangJiameng:
When there is no candidate, `ZeroDivisionError: float division by zero` is raised. Here the bug is fixed. 
However, it seems that the current devel branch can't make it through unittest. But it has nothing to do with this PR. 
```
FileExistsError: [Errno 17] File exists: '/root/dpgen/tests/data/abacus.in/Cu_ONCV_PBE-1.0.upf' -> 'cuw.fcc.01x01x01/00.place_ele/sys-0003-0001/Cu_ONCV_PBE-1.0.upf'
```


dpgen/simplify/simplify.py
```diff
@@ -408,7 +408,10 @@ def post_model_devi(iter_index, jdata, mdata):
     np.random.shuffle(idx)
     pick_idx = idx[:iter_pick_number]
     rest_idx = idx[iter_pick_number:]
-    dlog.info("total candidate {0:6d}   picked {1:6d} ({2:6.2f} %) rest {3:6d} ({4:6.2f} % )".format\
+    if(counter['candidate'] == 0) : 
+      dlog.info("no candidate")
+    else :
+      dlog.info("total candidate {0:6d}   picked {1:6d} ({2:6.2f} %) rest {3:6d} ({4:6.2f} % )".format\
```
njzjz, line 414: We use 4 spaces for indents

njzjz (approved):

PR #768, opened 2022-07-05
for sys_entire, set labeled=False

HuangJiameng:
Since DP-GEN always deal with Systems (instead of LabeledSystems) here,  labeled should be set as False

### DIFF
--- a/dpgen/simplify/simplify.py
+++ b/dpgen/simplify/simplify.py
@@ -218,7 +218,7 @@ def post_model_devi(iter_index, jdata, mdata):
     sys_failed = dpdata.MultiSystems()
     
     labeled = jdata.get("labeled", False)
-    sys_entire = dpdata.MultiSystems().from_deepmd_npy(os.path.join(work_path, rest_data_name + ".old"), labeled)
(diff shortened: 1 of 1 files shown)


njzjz:
Not always. `simplify` allows using `LabeledSystem` without recalculating the points.

There is a parameter `labeled` to control it. See line 44 and line 376.

dpgen/simplify/simplify.py
```diff
@@ -216,8 +216,9 @@ def post_model_devi(iter_index, jdata, mdata):
     sys_accurate = dpdata.MultiSystems()
     sys_candinate = dpdata.MultiSystems()
     sys_failed = dpdata.MultiSystems()
-
-    sys_entire = dpdata.MultiSystems().from_deepmd_npy(os.path.join(work_path, rest_data_name + ".old"))
+    
+    labeled = jdata.get("labeled", False)
+    sys_entire = dpdata.MultiSystems().from_deepmd_npy(os.path.join(work_path, rest_data_name + ".old"), labeled)
```
njzjz, line 221:
```suggestion
    sys_entire = dpdata.MultiSystems().from_deepmd_npy(os.path.join(work_path, rest_data_name + ".old"), labeled=labeled)
```

njzjz (changes requested):

njzjz (approved):

PR #769, opened 2022-07-06
fix ZeroDivisionError raised by run.py

HuangJiameng:
When tcount == 0, 'ZeroDivisionError: float division by zero' is raised. Here the bug is fixed.


njzjz (approved):

njzjz:
You may set a proper PR title, so we can know what the PR does without entering the PR.

HuangJiameng:
> You may set a proper PR title, so we can know what the PR does without entering the PR.

Maybe this one is more proper?

wanghan-iapcm (approved):

PR #783, opened 2022-07-07
Release DP-GEN v0.10.6

HuangJiameng:
Bohrium 1.1.1 need a mirror of the latest DP-GEN version (may be released as 0.10.6)


AnguseZhang:
@amcadmus Would this way work? Or we should create personal branch, merge master to devel, merge devel to master and release the new version? A little confused about the process to release.

PR #798, opened 2022-07-12
change data.init to relative path

HuangJiameng:
see [github.com/deepmodeling/dpgen/issues/797](url)


HuangJiameng:
I will check it.

PR #800, opened 2022-07-13
remove param start_pref_pf and limit_pref_pf

HuangJiameng:
start_pref_pf and limit_pref_pf are deprecated keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1064,9 +1064,7 @@ Here is an example of `param.json` for QM7 dataset:
             "start_pref_f": 1000,
             "limit_pref_f": 1,
             "start_pref_v": 0,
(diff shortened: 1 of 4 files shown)


njzjz:
They are not deprecated. See https://docs.deepmodeling.com/projects/deepmd/en/master/train/train-input.html#argument:loss[ener]/start_pref_pf

HuangJiameng:
I see. They are used in deepmd-kit. I haven't find them in the code of dpgen so I make a mistake.

njzjz:
Ok. However, I do notice that all these inputs are DeePMD-kit v1 format. They should be updated to v2 format.

Issue #794, opened 2022-07-11
[BUG] job.json is not found in post_fp_vasp process


HuangJiameng:
**Summary**

See [nohup.txt](https://github.com/deepmodeling/dpgen/files/9082205/nohup.txt)

**DPGEN Version**

dpgen v0.10.5-devel
(it is said that v0.10.3 can do simplify successfully)

**Expected Behavior**

job.json exists in /iter.000000/02.fp/task.*/

**Actual Behavior**

it doesn't exist there

**Further Info**
[INCAR.txt](https://github.com/deepmodeling/dpgen/files/9082223/INCAR.txt)
[machine_test_simplify.json.txt](https://github.com/deepmodeling/dpgen/files/9082224/machine_test_simplify.json.txt)
[nohup.out_bak_00-07.txt](https://github.com/deepmodeling/dpgen/files/9082225/nohup.out_bak_00-07.txt)
[record.dpgen.txt](https://github.com/deepmodeling/dpgen/files/9082226/record.dpgen.txt)
[simplify.json.txt](https://github.com/deepmodeling/dpgen/files/9082227/simplify.json.txt)
[dpdispatcher.log.txt](https://github.com/deepmodeling/dpgen/files/9082228/dpdispatcher.log.txt)
[dpgen.log.txt](https://github.com/deepmodeling/dpgen/files/9082229/dpgen.log.txt)

PR #803, opened 2022-07-15
fix post_fp_vasp and make_fp_configs for simplify

HuangJiameng:
When `job.json` is empty, it won't be generated. But post_fp_vasp will call it without checking if it exists. see [https://github.com/deepmodeling/dpgen/issues/794](issue #794)
Besides, there's something wrong with naming task directories. dpdata has to append all different systems into one set, which causes `RuntimeError('systems with inconsistent formula could not be append: %s v.s. %s' % (self.uniq_formula, system.uniq_formula))`. Now those problems are fixed.

### DIFF
--- a/dpgen/simplify/simplify.py
+++ b/dpgen/simplify/simplify.py
@@ -315,14 +315,16 @@ def make_fp_configs(iter_index, jdata):
     create_path(work_path)
     picked_data_path = os.path.join(iter_name, model_devi_name, picked_data_name)
     systems = get_multi_system(picked_data_path, jdata)
+    ii = 0
     jj = 0
     for system in systems:
         for subsys in system:
-            task_name = "task." + fp_task_fmt % (0, jj)
+            task_name = "task." + fp_task_fmt % (ii, jj)
             task_path = os.path.join(work_path, task_name)
             create_path(task_path)
             subsys.to('vasp/poscar', os.path.join(task_path, 'POSCAR'))
             jj += 1
+        ii += 1
 
 
 def make_fp_gaussian(iter_index, jdata):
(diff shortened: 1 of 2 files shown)


njzjz:
instead, we should change `LabeledSystem` to `MultiSystems`.

dpgen/simplify/simplify.py
```diff
@@ -315,14 +315,16 @@ def make_fp_configs(iter_index, jdata):
     create_path(work_path)
     picked_data_path = os.path.join(iter_name, model_devi_name, picked_data_name)
     systems = get_multi_system(picked_data_path, jdata)
+    ii = 0
     jj = 0
     for system in systems:
         for subsys in system:
-            task_name = "task." + fp_task_fmt % (0, jj)
+            task_name = "task." + fp_task_fmt % (ii, jj)
```
njzjz, line 322: Instead, I suggest using `MultiSystems` to `append` systems, like what dpgen2 does, so there will be no errors to add different systems.
https://github.com/deepmodeling/dpgen2/blob/008d1dfe91f3628a7039dca5d3af10c1b4bca506/dpgen2/op/collect_data.py#L67-L73

HuangJiameng:
But here has already used `MultiSystems`. 
https://github.com/deepmodeling/dpgen/blob/3c6803b4c37022fe474e496eaea7c90a6f700293/dpgen/simplify/simplify.py#L52-L53

njzjz:
here
https://github.com/deepmodeling/dpgen/blob/3c6803b4c37022fe474e496eaea7c90a6f700293/dpgen/generator/run.py#L3135-L3138

HuangJiameng:
Many operations are dependent on `LabeledSystem`. Maybe not to change it?

njzjz:
ok, I agree, although it's not a good behavior...

njzjz (approved):

AnguseZhang (approved):

PR #806, opened 2022-07-16
args: add `shuffle_poscar` to run jdata

njzjz:


wanghan-iapcm (approved):

dpgen/generator/arginfo.py
```diff
@@ -159,6 +160,7 @@ def model_devi_lmp_args() -> List[Argument]:
                  doc=doc_model_devi_nopbc),
         Argument("model_devi_activation_func", list, optional=True,
                  doc=doc_model_devi_activation_func),
+        Argument("shuffle_poscar", bool, optional=False, doc=doc_shuffle_poscar),
```
wanghan-iapcm, line 163: Better to set optional=True, default=False

wanghan-iapcm:
@njzjz any hint why the UTs fail...?

Issue #813, opened 2022-07-22
`load_ckpt` should be removed from README.md


HuangJiameng:
https://github.com/deepmodeling/dpgen/blob/devel/README.md?plain=1#L1079
![origin_img_v2_5ed7bc85-e36d-4c45-8873-5fbf1922437g](https://user-images.githubusercontent.com/105633685/180338819-9191cf86-a911-4d79-ac4a-95eda44de15c.jpg)
`load_ckpt` is not allowed now, it should be removed from README.md.

PR #817, opened 2022-07-22
Fix readme and qm7.json

HuangJiameng:
- correct the wrong spelling of "perturb" and "perturbation" in `README.md`
- remove "load_ckpt", see #813


njzjz:
Can you also remove `load_ckpt` from other examples?

HuangJiameng:
> Can you also remove `load_ckpt` from other examples?

I have updated 2 examples without specific `dp` version.

njzjz (approved):

PR #818, opened 2022-07-23
args: trust level for model_devi_job should be optional

njzjz:
The arguments #786 added should be optional.


njzjz:
@Cloudac7

wanghan-iapcm (approved):

Cloudac7:
With now `dargs` based argument checking, it raise `dargs.dargs.ArgumentKeyError` due to lack of `model_devi_f_trust_lo` when setting local `model_devi_f_trust_lo`. Should `Optional` be set to `True` for `model_devi_f_trust_lo` in `model_devi_lmp_args`?

PR #836, opened 2022-08-02
test: update abacus input parameters for test_make_fp.py

hongriTianqi:
update abacus input parameters for test_make_fp.py


wanghan-iapcm:
Could you please explain the change of the keywords? It that due the update of abacus version? Which version are the new keywords compatible to?

hongriTianqi:
> Could you please explain the change of the keywords? It that due the update of abacus version? Which version are the new keywords compatible to?

OK. Yes, this is due to the update of abacus version. The new keywords are compatible to abacus versions higher than 2.2, since there was important change of input keywords from 2.2 version of abacus. The changed keywords include:
dr2 => scf_thr
niter => scf_nmax
smearing => smearing_method
sigma => smearing_sigma
force => cal_force
stress => cal_stress
out_descriptor => deepks_out_labels
lmax_descriptor => deepks_descriptor_lmax
model_file => deepks_model

dpgen/generator/lib/abacus_scf.py
```diff
@@ -22,68 +22,97 @@ def make_abacus_scf_input(fp_params):
...
             assert(fp_params["deepks_descriptor_lmax"] >= 0),  "'deepks_descriptor_lmax' should be  a positive integer."
             ret += "deepks_descriptor_lmax %d\n" % fp_params["deepks_descriptor_lmax"]
         elif key == "deepks_scf":
+            if type(fp_params["deepks_scf"])==str:
+                fp_params["deepks_scf"] = int(eval(fp_params["deepks_scf"]))
             assert(fp_params["deepks_scf"] == 0  or fp_params["deepks_scf"] == 1), "'deepks_scf' should be either 0 or 1."
             ret += "deepks_scf %d\n" % fp_params["deepks_scf"]
         elif key == "deepks_model":
             ret += "deepks_model %s\n" % fp_params["deepks_model"]
-        elif key != "k_points": # "k_points key is used to generate KPT file."
```
Liu-RX, line 86: This `elif` is necessary. The `k_points` is also specified in `fp_params`, but can not be written in `INPUT` file.

Liu-RX (changes requested):

dpgen/generator/lib/abacus_scf.py
hongriTianqi: This `k_points` tag as been upgraded on the same level as 'user_fp_param'. Users may also use 'fp_kpt_file' to read KPT file now.

dpgen/generator/lib/abacus_scf.py
```diff
@@ -22,15 +22,23 @@ def make_abacus_scf_input(fp_params):
     ret += "calculation scf\n"
     for key in fp_params:
         if key == 'ntype':
+            if type(fp_params["ntype"])==str:
+                fp_params["ntype"] = int(fp_params["ntype"])
```
wanghan-iapcm, line 26: how about writting
```suggestion
            fp_params["ntype"] = int(fp_params["ntype"])
```
In case someone provide fp_params["ntype"] = 2.0

dpgen/generator/lib/abacus_scf.py
Liu-RX: Got that. I have no other questions.

Liu-RX (approved):

dpgen/generator/lib/abacus_scf.py
hongriTianqi: Got it. I will modify lines having the same problem.

hongriTianqi:
The current test error seems not caused by this PR, since I installed the devel branch without this PR locally and got the same error.

njzjz:
The error is fixed by #839.

hongriTianqi:
> The error is fixed by #839.

Thanks!

dpgen/generator/run.py
```diff
@@ -3022,7 +3063,35 @@ def run_fp (iter_index,
...
+                    forward_files += ["KPT"]
+                else:
+                    if 'k_points' in jdata.keys() :
+                        forward_files += ["KPT"]
+                    elif 'fp_kpt_file' in jdata.keys():
+                        forward_files += ["KPT"]
+                    else:
+                        raise RuntimeError("Cannot find any k-points information")
+            else:
+                forward_files += ["KPT"]
```
wanghan-iapcm, line 3092: In all cases `forward_files += ["KPT"]`, why not simply write `forward_files = ["INPUT", "STRU", "KPT"]` ?

The validity of the kpoint information has been checked by `make_fp_abacus_scf`

hongriTianqi: Thanks for pointing this. Revised.

wanghan-iapcm (approved):

PR #840, opened 2022-08-04
add simplify example of MAPbI3-scan based on lebesgue

HuangJiameng:
Add an example for `dpgen simplify`. `data` contains a simplistic data set based on MAPbI3-scan case. `vasp_input` is an empty dir to remind that POTCAR files should be prepared. `simplify_example` is the work path, which contains `INCAR` and templates for PARAM and MACHINE files.


HuangJiameng:
These POTCAR files should be prepared by the user.
```
└── vasp_inputs
    ├── POTCAR_C
    ├── POTCAR_H
    ├── POTCAR_I
    ├── POTCAR_N
    └── POTCAR_Pb_d
```

examples/simplify-MAPbI3-scan-lebesgue/simplify_example/simplify.json
```diff
@@ -0,0 +1,98 @@
+{
```
njzjz, line 1: Could you add the filename after the following line, so we do not need to manually check this file.
https://github.com/deepmodeling/dpgen/blob/db6c8f111905481659720dd236f419e33308fcf3/tests/test_check_examples.py#L30

wanghan-iapcm (commented):
Is it possible to reduce the size of the dataset? The example is only for demonstrative purpose.

HuangJiameng:
> Is it possible to reduce the size of the dataset? The example is only for demonstrative purpose.

It has already been reduced from 14 sets to 3.

wanghan-iapcm:
> > Is it possible to reduce the size of the dataset? The example is only for demonstrative purpose.
> 
> It has already been reduced from 15 sets to 3.

The size of examples/simplify-MAPbI3-scan-lebesgue/data/init.000/ is over 5M. The other two datasets are ~0.8M. 

We want the source code light-weighted. Please reduce the size of the dataset. 

In principle you only need to provide data for a few iteration. If one iteration takes too many data, you may reduce the `xxx_pick_number` parameters.

njzjz:
Conflicts need to be resolved.

HuangJiameng:
I will create a new branch and make PR again after these changes.

dpgen/simplify/arginfo.py
```diff
@@ -25,6 +25,7 @@ def general_simplify_arginfo() -> Argument:
     doc_iter_pick_number = "The number of pick data in each iteration."
     doc_model_devi_f_trust_lo = "The lower bound of forces for the selection for the model deviation."
     doc_model_devi_f_trust_hi = "The higher bound of forces for the selection for the model deviation."
+    doc_shuffle_poscar = 'Shuffle atoms of each frame before running simulations. The purpose is to sample the element occupation of alloys.'
```
njzjz, line 28: What is "simulation" in the simplify workflow?

njzjz: Does this option take effect?

HuangJiameng: No. It is set as false in the PARAM I get from the real case. I will remove it.

njzjz (approved):

Issue #333, opened 2020-12-05
Issue with LAMMPS jobs in DPGEN procedure


cesaremalosso:
Hi,
I'm trying to use your package in order to train a neural network potential to simulate water.
The procedure seems to proceed correctly till it's time for LAMMPS to generate new configurations using the just trained potentials. The `dpgen.log` gives me back the following text. It seems like the stage 4 of the iteration (`run_model_devi`), which should consist in the submissions of the LAMMPS job, does not happen. Anyway, no errors are present.

```
2020-12-05 18:31:13,398 - INFO : start running
2020-12-05 18:31:13,401 - INFO : =============================iter.000000==============================
2020-12-05 18:31:13,401 - INFO : -------------------------iter.000000 task 00--------------------------
2020-12-05 18:31:13,419 - INFO : -------------------------iter.000000 task 01--------------------------
2020-12-05 18:31:13,615 - INFO : new submission of e26935fc-bf92-41b8-89f0-608d0cf6a83d for chunk 8aefb06c426e07a0a671a1e2488b4858d694a730
2020-12-05 18:31:13,707 - INFO : new submission of 31803ae5-570c-42ba-9543-48a4fa40bcc4 for chunk e193a01ecf8d30ad0affefd332ce934e32ffce72
2020-12-05 18:31:13,828 - INFO : new submission of 4a9babda-ba80-4558-bc7f-d686547281fc for chunk 6fc978af728d43c59faa400d5f6e0471ac850d4c
2020-12-05 18:31:14,009 - INFO : new submission of 07831ff6-e1fd-4894-a8f2-5d908b206c5d for chunk 221407c03ae5c73109cce71d27e24637824f3333
2020-12-05 18:32:14,187 - INFO : job e26935fc-bf92-41b8-89f0-608d0cf6a83d finished
2020-12-05 18:33:14,413 - INFO : job 31803ae5-570c-42ba-9543-48a4fa40bcc4 finished
2020-12-05 18:34:14,588 - INFO : job 4a9babda-ba80-4558-bc7f-d686547281fc finished
2020-12-05 18:35:14,748 - INFO : job 07831ff6-e1fd-4894-a8f2-5d908b206c5d finished
2020-12-05 18:35:14,800 - INFO : -------------------------iter.000000 task 02--------------------------
2020-12-05 18:35:14,803 - INFO : -------------------------iter.000000 task 03--------------------------
[...]

PR #854, opened 2022-08-09
Check run_tasks before submission in run_md_model_devi

HuangJiameng:
`run_tasks` for `model_devi` step should not be empty. see #333 
a warning will be given when `run_tasks` is abnormal

### DIFF
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -1578,6 +1578,8 @@ def run_md_model_devi (iter_index,
                         errlog = 'model_devi.log')
 
     elif LooseVersion(api_version) >= LooseVersion('1.0'):
+        if(len(run_tasks) == 0): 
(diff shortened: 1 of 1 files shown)


dpgen/generator/run.py
```diff
@@ -1578,6 +1578,8 @@ def run_md_model_devi (iter_index,
                         errlog = 'model_devi.log')
 
     elif LooseVersion(api_version) >= LooseVersion('1.0'):
+        if(len(run_tasks) == 0): 
+            warnings.warn("run_tasks for model_devi should not be empty! Please check your files.")
```
njzjz, line 1582: Shall we raise `RuntimeError` instead? I don't think the workflow can be continued when nothing is running.

njzjz, line 1582: I don't think it should be inside `elif` block.

njzjz (approved):

AnguseZhang (approved):

PR #861, opened 2022-08-11
Change the location of “export ” in Slurm.py

Zhang-pchao:
In more cases, shoule the location of "export" be placed in front of "source"?
For example, it applies to the following situations (otherwise my_package cannot be found):

_export PATH="/home_path/anaconda/envs/my_package/bin:$PATH"
source activate my_package_


wanghan-iapcm:
The dispatcher module in dpgen is deprecated and will be removed in the future. Now dpgen uses [dpdispatcher](https://github.com/deepmodeling/dpdispatcher). Please contribute to dpdispatcher.

njzjz:
Old dispatcher has been removed from DP-GEN. Please contribute to the new dpdispatcher instead.

PR #864, opened 2022-08-12
update run/dp2.x-lammps-cp2k example

HuangJiameng:


HuangJiameng:
`train_param` and `ratio_failed` both work in `dpgen run`, but why they can't pass the check?

njzjz:
I don't think `train_param` still works. You can add `ratio_failed` to `arginfo`.

njzjz (approved):

PR #867, opened 2022-08-12
Automatically set the relative atomic mass according to "type_map"

panxiang126:
A new function `get_atomic_masses` was written to get the relative atomic mass (data from: [Atomic weights of the elements 2013 (IUPAC Technical Report)](https://www.degruyter.com/document/doi/10.1515/pac-2015-0305/html)). So that there is no need to specify the `mass_map` manually in `param.json`.


dpgen/generator/run.py
```diff
@@ -3733,6 +3733,20 @@ def run_iter (param_file, machine_file) :
                 raise RuntimeError ("unknown task %d, something wrong" % jj)
             record_iter (record, ii, jj)
 
+        
+def get_atomic_masses(chemical_symbol):
+  chemical_symbols = ['H', 'He', 'Li', 'Be', 'B', 'C', 'N', 'O', 'F', 'Ne', 'Na', 'Mg', 'Al', 'Si', 'P', 'S', 'Cl', 'Ar', 'K', 'Ca', 'Sc', 'Ti', 'V', 'Cr', 'Mn', 'Fe', 'Co', 'Ni', 'Cu', 'Zn', 'Ga', 'Ge', 'As', 'Se', 'Br', 'Kr', 'Rb', 'Sr', 'Y', 'Zr', 'Nb', 'Mo', 'Tc', 'Ru', 'Rh', 'Pd', 'Ag', 'Cd', 'In', 'Sn', 'Sb', 'Te', 'I', 'Xe', 'Cs', 'Ba', 'La', 'Ce', 'Pr', 'Nd', 'Pm', 'Sm', 'Eu', 'Gd', 'Tb', 'Dy', 'Ho', 'Er', 'Tm', 'Yb', 'Lu', 'Hf', 'Ta', 'W', 'Re', 'Os', 'Ir', 'Pt', 'Au', 'Hg', 'Tl', 'Pb', 'Bi', 'Po', 'At', 'Rn', 'Fr', 'Ra', 'Ac', 'Th', 'Pa', 'U', 'Np', 'Pu', 'Am', 'Cm', 'Bk', 'Cf', 'Es', 'Fm', 'Md', 'No', 'Lr', 'Rf', 'Db', 'Sg', 'Bh', 'Hs', 'Mt', 'Ds', 'Rg', 'Cn', 'Nh', 'Fl', 'Mc', 'Lv', 'Ts', 'Og']
+
+  # doi:10.1515/pac-2015-0305
```
njzjz, line 3740: Why not use the latest data?
https://doi.org/10.1515/pac-2019-0603

dpgen/generator/run.py
```diff
@@ -1115,7 +1115,7 @@ def _make_model_devi_native(iter_index, jdata, mdata, conf_systems):
     model_devi_taup = 0.5
     if 'model_devi_taup' in jdata :
         model_devi_taup = jdata['model_devi_taup']
-    mass_map = jdata['mass_map']
+    mass_map = [get_atomic_masses(i) for i in jdata['type_map']]
```
njzjz, line 1118: This should not override the default config, as one may want to simulate isotopes.

njzjz (changes requested):

wanghan-iapcm (commented):
The `dpgen` and `dpdata` do not restrict the `atom_names` to chemical symbols. This PR will break the convention.

dpgen/generator/run.py
panxiang126: updated.

dpgen/generator/run.py
panxiang126: updated.

Please check if `jdata['mass_map'] is None` is used correctly

dpgen/generator/run.py
panxiang126: I can not fully understand [wanghan-iapcm](https://github.com/wanghan-iapcm)'s comment below.  I have rewritten the `get_atomic_masses ` function, and now it can be called by three types, _i. e._ `element_names`, `chemical_symbols` and `atomic_number`. I am not sure if this will remove his/her concern.

dpgen/generator/run.py
```diff
@@ -3733,6 +3734,34 @@ def run_iter (param_file, machine_file) :
...
...n', 'Potassium', 'Calcium', 'Scandium', 'Titanium', 'Vanadium', 'Chromium', 'Manganese', 'Iron', 'Cobalt', 'Nickel', 'Copper', 'Zinc', 'Gallium', 'Germanium', 'Arsenic', 'Selenium', 'Bromine', 'Krypton', 'Rubidium', 'Strontium', 'Yttrium', 'Zirconium', 'Niobium', 'Molybdenum', 'Technetium', 'Ruthenium', 'Rhodium', 'Palladium', 'Silver', 'Cadmium', 'Indium', 'Tin', 'Antimony', 'Tellurium', 'Iodine', 'Xenon', 'Caesium', 'Barium', 'Lanthanum', 'Cerium', 'Praseodymium', 'Neodymium', 'Promethium', 'Samarium', 'Europium', 'Gadolinium', 'Terbium', 'Dysprosium', 'Holmium', 'Erbium', 'Thulium', 'Ytterbium', 'Lutetium', 'Hafnium', 'Tantalum', 'Tungsten', 'Rhenium', 'Osmium', 'Iridium', 'Platinum', 'Gold', 'Mercury', 'Thallium', 'Lead', 'Bismuth', 'Polonium', 'Astatine', 'Radon', 'Francium', 'Radium', 'Actinium', 'Thorium', 'Protactinium', 'Uranium', 'Neptunium', 'Plutonium', 'Americium', 'Curium', 'Berkelium', 'Californium', 'Einsteinium', 'Fermium', 'Mendelevium', 'Nobelium', 'Lawrencium', 'Rutherfordium', 'Dubnium', 'Seaborgium', 'Bohrium', 'Hassium', 'Meitnerium', 'Darmastadtium', 'Roentgenium', 'Copernicium', 'Nihonium', 'Flerovium', 'Moscovium', 'Livermorium', 'Tennessine', 'Oganesson']
```
njzjz, line 3739: We use 4 spaces for indent.

njzjz (changes requested):
Please update the following documentation to reflect your changes.

https://github.com/deepmodeling/dpgen/blob/741f4cda3e51495df5e812abcbbb1f0296f85698/dpgen/generator/arginfo.py#L20

dpgen/generator/run.py
wanghan-iapcm: I mean atom name can be ANY string, not necessary element_names or chemical_symbols. 
For example it is valid to write `"type_map" : ["foo", "bar"]`.

dpgen/generator/run.py
panxiang126: updated.

dpgen/generator/arginfo.py
```diff
@@ -17,15 +17,15 @@ def run_mdata_arginfo() -> Argument:
...
+    doc_mass_map = 'Standard atomic weights. You can set a list of atomic masses corresponding to type_map, or just set "auto"'
     doc_use_ele_temp = 'Currently only support fp_style vasp. \n\n\
 - 0: no electron temperature. \n\n\
 - 1: eletron temperature as frame parameter. \n\n\
 - 2: electron temperature as atom parameter.'
 
     return [
         Argument("type_map", list, optional=False, doc=doc_type_map),
-        Argument("mass_map", list, optional=False, doc=doc_mass_map),
+        Argument("mass_map", [str, list], optional=True, default="auto", doc=doc_mass_map),
```
njzjz, line 28: Note: as the arguments are still being built, at present the default value will not be automatically applied, but only shown in the documentation.

njzjz (approved):

dpgen/generator/run.py
njzjz: I agree with @wanghan-iapcm. This should be documented that `auto` only supports the standard atoms. (For example, `OW` will not be supported)

panxiang126: I updated the documentation.

njzjz (approved):

njzjz:
@AnguseZhang you should squash merge...

dpgen/generator/run.py
```diff
@@ -968,7 +968,7 @@ def _make_model_devi_revmat(iter_index, jdata, mdata, conf_systems):
     sys_idx = expand_idx(cur_job['sys_idx'])
     if (len(sys_idx) != len(list(set(sys_idx)))) :
         raise RuntimeError("system index should be uniq")
-    mass_map = jdata['mass_map']
+    mass_map = [get_atomic_masses(i) for i in jdata['type_map']] if jdata['mass_map'] == "auto" else jdata['mass_map']
```
wanghan-iapcm, line 971: You may want to write a function for automatic building of mass_map rather than repeating the same line for multiple times.

wanghan-iapcm:
@AnguseZhang I am against the merging of this PR.

AnguseZhang:
> @AnguseZhang I am against the merging of this PR.

@wanghan-iapcm Reverted in #881

PR #869, opened 2022-08-13
docs: add authors section for credits

njzjz:


njzjz:
https://deepmodeling--869.org.readthedocs.build/projects/dpgen/en/869/credits.html

PR #870, opened 2022-08-14
make more examples pass argument tests

njzjz:


dpgen/generator/arginfo.py
```diff
@@ -321,12 +321,14 @@ def fp_style_siesta_args() -> List[Argument]:
...
     doc_NumberPulay = 'Controls the Pulay convergence accelerator.'
+    doc_fp_pp_path = 'Directory of psuedo-potential or numerical orbital files to be used for 02.fp exists.'
+    doc_fp_pp_files = 'Psuedo-potential file to be used for 02.fp. Note that the order of elements should correspond to the order in type_map.'
 
     args = [
         Argument("ecut", int, optional=False, doc=doc_ecut),
         Argument("ediff", float, optional=False, doc=doc_ediff),
-        Argument("kspacing", float, optional=False, doc=doc_kspacing),
-        Argument("mixingweight", float, optional=False, doc=doc_mixingweight),
+        Argument("kspacing", [float, int], optional=False, doc=doc_kspacing),
```
wanghan-iapcm, line 330: Why `kspacing` can be `int`?

njzjz: It's not clear to me if the following example is a typo or not. I am not familiar with siesta.

https://github.com/deepmodeling/dpgen/blob/8dea29ef125f66be9641afe5ac4970433a9c9ce1/examples/run/deprecated/dp0.12-lammps-siesta/dp-lammps-siesta/CH4/param_CH4.json#L129

njzjz: cc @Silvia-liu

HuangJiameng:
Since these examples are modified to pass the current argument tests, are they still available for `dp1.x`?

njzjz:
I don't think there are any differences in DP-GEN parameters between `dp1.x` and `dp2.x`.

However, it should be a good idea to upgrade all DeePMD-kit parameters from 1.x to 2.x.

HuangJiameng (approved):

dpgen/generator/arginfo.py
wanghan-iapcm: It should be `"kspacing" : 1,`

wanghan-iapcm: It should be `"kspacing" : 1.0,`

wanghan-iapcm (approved):

PR #873, opened 2022-08-15
Example for dp2.x_lmp_gaussian

ziqi-hu:
An example.


examples/run/dp2.x-lammps-gaussian/data/deepmd/data.000/H0C0N2/box.raw
```diff
@@ -0,0 +1,45 @@
+1.000000000000000000e+02 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 1.000000000000000000e+02 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 1.000000000000000000e+02
```
njzjz, line 1: Data is too much for an example.

njzjz:
After #870 and #874 are merged, this PR should follow the same way to check arguments. We don't want to manually review it.

PR #874, opened 2022-08-15
add tests to check machine files

njzjz:


njzjz:
@HuangJiameng, @hongriTianqi, @felix5572: I modified some of the machine files you just added to make the tests pass.

HuangJiameng (approved):

examples/run/dp2.x-lammps-ABACUS-lcao/fcc-al/machine.json
```diff
@@ -25,18 +25,17 @@
...
                 "image_name":"",
-                "on_demand":0,
-                "job_type":""
+                "on_demand":0
             }
         }
       },
       "resources": {
-        "local_root":"./",
+        "batch_type": "Lebesgue",
```
hongriTianqi, line 33: Why delete "local_root", and add "batch_type"? Shouldn't "batch_type" be added in "machine" block?

njzjz: I don't see why you add `local_root`. It's never a parameter of `resources`.

njzjz: `batch_type` is used for checking as different types have different parameters.

hongriTianqi: > I don't see why you added `local_root`. It's never a parameter of `resources`.

This machine.json file was obtained from users in AISI, local_root was set from beginning. I will check with them.

hongriTianqi: > `batch_type` is used for checking as different types have different parameters.

`batch_type` has been set in the `machine` block, is it necessary to repeat this information in `resources` block?

hongriTianqi (approved):

examples/run/dp2.x-lammps-ABACUS-lcao/fcc-al/machine.json
njzjz: @y1xiaoc do you think it is possible to use an argument in another block as the variant key?

y1xiaoc: > @y1xiaoc do you think it is possible to use an argument in another block as the variant key?

In principle yes but I may need to redesign current dargs code structure. Are these two keys always have to be the same? Currently one way to enforce that is to set the `extra_check` parameter in their comment parent argument.

wanghan-iapcm (approved):

PR #815, opened 2022-07-22
fix bug of matrix multiply in `create_disturbs_abacus_dev`

pxlxingliang:

PR #878, opened 2022-08-16
DPGEN with constrained structures

SwiftChen97:
Add a new feature which can be used to generate constrained slab models automatically by specifying keywords "total_layers" and "fix_layers" in param.json (dpgen init_surf PARAM [MACHINE]), perturbating the atoms which is not fixed.

Co-Authored-By: Cyanbeacon <86344464+Cyanbeacon@users.noreply.github.com>


njzjz:
Please resolve conflicts.

wanghan-iapcm:
Please check if I understood correctly: This PR add constrains to POSCARS, so one may conduct the constrained vasp AIMD simulation of surface structures. 
In general, one does NOT need to do AIMD simulations for surface structures. Starting from a relatively good DP model for bulk structures, then explore the surface structures with DPMD is recommended, because it is much more efficient than AIMD.

SwiftChen97:
> Please check if I understood correctly: This PR add constrains to POSCARS, so one may conduct the constrained vasp AIMD simulation of surface structures. In general, one does NOT need to do AIMD simulations for surface structures. Starting from a relatively good DP model for bulk structures, then explore the surface structures with DPMD is recommended, because it is much more efficient than AIMD.

Yes, your understanding is correct. This PR can generate constrained surface structures. The perturbations will only be added to the atoms which haven't been constrained. It seems that this PR is not necessary for training potential functions of surface models based on what you said.

SwiftChen97:
> Please check if I understood correctly: This PR add constrains to POSCARS, so one may conduct the constrained vasp AIMD simulation of surface structures. In general, one does NOT need to do AIMD simulations for surface structures. Starting from a relatively good DP model for bulk structures, then explore the surface structures with DPMD is recommended, because it is much more efficient than AIMD.

Maybe It can be used to generate constrained surface models with perturbations for other explorations.

dpgen/data/tools/create_random_disturb.py
```diff
@@ -214,6 +223,7 @@ def create_disturbs_abacus_dev(fin, nfile, dmax=1.0, etmax=0.1, ofmt="abacus", d
         stru_d['cells'] = cell_new
 
         convert_mat = np.linalg.inv(cell).dot(cell_new)
+        convert_mat = np.linalg.inv(cell)*cell_new
```
wanghan-iapcm, line 226: This was a bug fixed #815 .

It seems that at some point you did not track the changes with git. I suspect there are other unwanted changes in this PR. 

I suggest you checking out a new branch based on the latest `devel` branch, applying **your changes** on it and then resubmitting the PR.

wanghan-iapcm (changes requested):

dpgen/data/tools/create_random_disturb.py
SwiftChen97: Ok, thanks for your suggestion. I will track the latest **devel** branch and resubmit the PR.

Issue #737, opened 2022-05-25
[BUG] `post_fp_check_fail` does not work for dpdispatcher


njzjz:
**Summary**

`tag_failure` checks the number of the failure tag, but dpdispatcher doesn't have this file.

https://github.com/deepmodeling/dpgen/blob/0767dce7cad29367edb2e4a55fd0d8724dbda642/dpgen/generator/run.py#L2717

The alternative way is to check the number of collected data. The failure ratio is $1-\frac{N_\text{data}}{N_\text{task}}$.

**DPGEN Version and Platform**

devel

PR #882, opened 2022-08-17
fix: check the number of collected data in `post_fp_check_fail`

HuangJiameng:
see #737 
dpdispatcher use flag_if_job_task_fail to mark the failed jobs, so `post_fp_check_fail` can be used before checking the frames. If we use the alternative way the issue mentioned, we should consider different representative outputs according to `fp_style`. I am wondering if it is a repeat of the following frame checks. However, since `flag_if_job_task_fail` will be marked as True if one task in the group is failed, I am afraid that `rfail` could be high when only a few tasks fail. I'd like to ask for some suggestions.

### DIFF
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -3181,7 +3181,7 @@ def post_fp_check_fail(iter_index,
     sys_data.sort()
     nframe = 0
     for ii in sys_data :
-        sys = dpdata.LabeledSystem().from_deepmd_raw(ii, type_map = jdata['type_map'])
+        sys = dpdata.LabeledSystem(ii, fmt = 'deepmd/raw')
         nframe += len(sys)
     nfail = ntask - nframe
 
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -3165,23 +3165,36 @@ def run_fp (iter_index,
 def post_fp_check_fail(iter_index,
                        jdata,
                        rfailed = None) :
+
     ratio_failed =  rfailed if rfailed else jdata.get('ratio_failed',0.05)
     iter_name = make_iter_name(iter_index)
     work_path = os.path.join(iter_name, fp_name)
-
-    # check fail according to _flag_if_job_task_fail, which will be 1 if any task in the group fails.
-    fp_flags = glob.glob(os.path.join(work_path, '*_flag_if_job_task_fail'))
-    if(len(fp_flags) == 0) :
+    fp_tasks = glob.glob(os.path.join(work_path, 'task.*'))
+    fp_tasks.sort()
+    if len(fp_tasks) == 0 :
         return
-    njob = len(fp_flags)
+    ntask = len(fp_tasks)
     nfail = 0
-    for ii in fp_flags:
-        with open(ii) as fp:
-            fail_flag = fp.read()
-            if fail_flag :
-                nfail += 1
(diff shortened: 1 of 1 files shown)


njzjz:
This is not correct as dpdispatcher does not allow some command to fail. It will retry and raise `RuntimeError`. If you want to allow some command to return a non-zero exit code, you can consider `some_command || touch some_flag` and then count `some_flag` instead.

HuangJiameng:
How about counting `*_task_tag_finished`? Each task has `task_tag_finished`. $rfail = 1 - \frac{nfinished}{ntasks}$ 
Creating a new tag seems cumbersome.

njzjz:
Dpdispatcher expects all tasks finished.

HuangJiameng:
Then `post_fp_check_fail` seems useless? Is it reasonable to discard it? It doesn't work now in fact.

njzjz:
It is used with the old dispatcher.

njzjz:
Also, I think the tag will not be backward unless it is added into the list of backward files.

HuangJiameng:
Since dpdispatcher will raise an error if any task fails, while the ratio of fail frames shows available data, shall we deprecate `post_fp_check_fail`?

njzjz:
One can use `some_command || :` in the command to allow `some_command` to fail (but the whole command still successes). And dpdata will skip non-coverage data. After new data is collected, we can count the number of data. The failed ratio can be calculated by as what I mentioned in #737.

HuangJiameng:
check fail according to the number of collected data when using dpdispatcher (api_version >= 1.0)

dpgen/generator/run.py
```diff
@@ -3165,20 +3165,26 @@ def run_fp (iter_index,
...
     ntask = len(fp_tasks)
-    nfail = len(fp_failed_tasks)
+    nfail = 0
+
+    # check fail according to the number of collected data
+    sys_data = glob.glob(os.path.join(work_path, "data.*"))
+    sys_data.sort()
+    nframe = 0
+    for ii in sys_data :
+        sys = dpdata.LabeledSystem(ii, fmt = 'deepmd/raw')
```
njzjz, line 3184:
```suggestion
        sys = dpdata.LabeledSystem(ii, fmt = 'deepmd/npy')
```

It looks like some styles don't output raw files.

HuangJiameng: Thanks for suggestion. Still, I am wondering how to fix unit tests. Orz

njzjz:
It looks like the failed unit test outputs a `MultiSystems` directory. Your method does not cover this situation. You may use `dpgen.util.expand_sys_str`.

dpgen/generator/run.py
```diff
@@ -3165,20 +3165,28 @@ def run_fp (iter_index,
...
+
+    # check fail according to the number of collected data
+    sys_data = glob.glob(os.path.join(work_path, "data.*"))
+    sys_data.sort()
+    nframe = 0
+    for ii in sys_data :
+        sys_paths = expand_sys_str(ii)
+        for single_sys in sys_paths:
+            sys = dpdata.LabeledSystem(os.path.join(single_sys), fmt = 'deepmd/npy')
+        nframe += len(sys)
```
njzjz, line 3187: This line should be in the `for` loop.

njzjz (approved):
LGTM

AnguseZhang (approved):

PR #892, opened 2022-08-19
DPGEN with constrained structures (new)

ShaneLogic:
To generate the constrained surface models proposed at discussion [#877](https://github.com/deepmodeling/dpgen/discussions/877)

Co-Authored-By: Cyanbeacon <86344464+Cyanbeacon@users.noreply.github.com>

### DIFF
--- a/dpgen/data/fix_layer.py
+++ b/dpgen/data/fix_layer.py
@@ -1,0 +1,50 @@
+import numpy as np
+
+def layer(jdata, poscar_in, poscar_out):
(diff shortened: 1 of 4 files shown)


wanghan-iapcm (changes requested):
Thank you very much for the nice contribution. Could you please:
- explanation how to use the atom constrain in the document
- an illustrative example would be welcome
- add UTs that make sure the code is working in an expected way.

PR #909, opened 2022-08-23
Support forward files for init_surf

HuangJiameng:
Support user use "forward_files" in machine parameters to upload other files besides POSCAR, INCAR, POTCAR. (forwarded files for each task)

### DIFF
--- a/dpgen/data/surf.py
+++ b/dpgen/data/surf.py
@@ -548,6 +548,13 @@ def run_vasp_relax(jdata, mdata):
     forward_files = ["POSCAR", "INCAR", "POTCAR"]
     backward_files = ["OUTCAR","CONTCAR"]
     forward_common_files = []
+    more_forward_files_path_list = mdata['fp'].get('forward_files', []) 
+    more_forward_files = []
+    for ii in more_forward_files_path_list :
+        os_path_ii = os.path.join(ii)
+        more_forward_files.append(os_path_ii)
+        shutil.copy(os_path_ii, work_dir) 
+        forward_files.append(os.path.basename(os_path_ii) )
     #if 'cvasp' in mdata['fp_resources']:
     #    if mdata['fp_resources']['cvasp']:
     #        forward_common_files=['cvasp.py']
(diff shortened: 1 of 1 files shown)


njzjz:
Isn't it `user_forward_files`?

HuangJiameng:
It is `forward_files` in machine parameters. https://docs.deepmodeling.com/projects/dpdispatcher/en/latest/task.html#argument:task/forward_files 
Am I understanding it wrong？

njzjz:
I think it's this feature:
https://github.com/deepmodeling/dpgen/blob/5dc7c1b3380c78df6c895bb956c94c283e2c9f56/dpgen/generator/run.py#L580-L582

Indeed we should not maintain it everywhere... Create a method instead.

HuangJiameng:
> I think it's this feature:
> 
> https://github.com/deepmodeling/dpgen/blob/5dc7c1b3380c78df6c895bb956c94c283e2c9f56/dpgen/generator/run.py#L580-L582
> 
> Indeed we should not maintain it everywhere... Create a method instead.

In contrast, I think it is more reasonable to put symlinks in the task path for paths in "forward_files". Otherwise the user should pile everything in the main directory. Besides, should we keep the same parameter name "forward_files" with dpdispatcher?

njzjz:
> > I think it's this feature:
> > https://github.com/deepmodeling/dpgen/blob/5dc7c1b3380c78df6c895bb956c94c283e2c9f56/dpgen/generator/run.py#L580-L582
> > 
> > Indeed we should not maintain it everywhere... Create a method instead.
> 
> In contrast, I think it is more reasonable to put symlinks in the task path for paths in "forward_files". Otherwise the user should pile everything in the main directory. Besides, should we keep the same parameter name "forward_files" with dpdispatcher?

I don't understand what you mean. The old codes also make symbolic links.

Making an alias is fine, but the old name should be kept for compatibility.

dpgen/data/surf.py
```diff
@@ -548,6 +549,13 @@ def run_vasp_relax(jdata, mdata):
     forward_files = ["POSCAR", "INCAR", "POTCAR"]
     backward_files = ["OUTCAR","CONTCAR"]
     forward_common_files = []
+    work_path_list = glob.glob(os.path.join(work_dir, "surf-*"))
+    task_format = {"fp" : "sys-*"}
+    for work_path in work_path_list :
+        symlink_user_forward_files(mdata=mdata, task_type="fp", work_path=work_path, task_format=task_format)
```
njzjz, line 555: Why not directly set format as `surf-*/sys-*`

HuangJiameng: Because `surf-*` seems to align with the meaning of `work_path`.

njzjz (approved):

PR #910, opened 2022-08-23
fix: equi_stress should have the same unit as stress

hongriTianqi:
equi_stress should have same unit as stress in elastic constant calculation.


kevinwenminion:
Thanks for carefully checking the code. This indeed needs to be fixed. Fortunately, it won't influence much on the elastic constants if the relaxation process is done correctly and the equilibrium stress is very small.

kevinwenminion (approved):

PR #917, opened 2022-08-25
Update run.py | Check if outcar exists.

ziqi-hu:
Add a error when OUTCAR does not exist in 02.fp/task.*.
When the vasp software is not running normally (such as mpirun commend not found), dpgen run will run normally.  I think the user should be notified of an error at this point.

Signed-off-by: ziqi-hu <51811006+ziqi-hu@users.noreply.github.com>

### DIFF
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -3218,7 +3218,7 @@ def post_fp_vasp (iter_index,
         sys_outcars = glob.glob(os.path.join(work_path, "task.%s.*/OUTCAR"%ss))
         sys_outcars.sort()
         if len(sys_outcars) == 0:
-                raise RuntimeError("OUTCAR does not exist! please check fp_tasks.")
+            raise RuntimeError("OUTCAR does not exist! please check fp_tasks.")
         tcount += len(sys_outcars)
         all_sys = None
         all_te = []
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -3217,6 +3217,8 @@ def post_fp_vasp (iter_index,
     for ss in system_index :
         sys_outcars = glob.glob(os.path.join(work_path, "task.%s.*/OUTCAR"%ss))
         sys_outcars.sort()
+        if len(sys_outcars) == 0:
+                raise RuntimeError("OUTCAR does not exist! please check fp_tasks.")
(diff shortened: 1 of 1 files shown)


HuangJiameng:
https://github.com/deepmodeling/dpgen/blob/devel/dpgen/generator/run.py#L3235-L3249

If an OUTCAR is unavailable, we will not raise error directly. Instead, we will count them into rfail. If rfail is higher than expected, RuntimeError "find too many unsuccessfully terminated jobs. Too many FP tasks are not converged. Please check your input parameters (e.g. INCAR) or configuration (e.g. POSCAR) in directories \'iter.*.*/02.fp/task.*.*/.\'" will be raised. Maybe you can change "find too many unsuccessfully terminated jobs" into "find too many unsuccessful OUTCAR!" Do you think it is clearer?

dpgen/generator/run.py
```diff
@@ -3217,6 +3217,8 @@ def post_fp_vasp (iter_index,
     for ss in system_index :
         sys_outcars = glob.glob(os.path.join(work_path, "task.%s.*/OUTCAR"%ss))
         sys_outcars.sort()
+        if len(sys_outcars) == 0:
+                raise RuntimeError("OUTCAR does not exist! please check fp_tasks.")
```
njzjz, line 3221: Note that only 4 spaces should be used as indents.

njzjz:
> https://github.com/deepmodeling/dpgen/blob/devel/dpgen/generator/run.py#L3235-L3249
> 
> If an OUTCAR is unavailable, we will not raise error directly. Instead, we will count them into rfail. If rfail is higher than expected, RuntimeError "find too many unsuccessfully terminated jobs. Too many FP tasks are not converged. Please check your input parameters (e.g. INCAR) or configuration (e.g. POSCAR) in directories 'iter._._/02.fp/task._._/.'" will be raised. Maybe you can change "find too many unsuccessfully terminated jobs" into "find too many unsuccessful OUTCAR!" Do you think it is clearer?

`len(sys_outcars)` is the number of `OUTCAR` files. It does not check whether it is coverage.

njzjz (commented):
This PR is not correct. We do not require every system has a task to calculate as it may be accurate enough.

PR #942, opened 2022-09-10
Dependency: Install pymatgen-analysis-defect for defect-related functionality

HuangJiameng:
- add `pymatgen-analysis-defect` and `pybind11` in `setup.py`.
- add reminder for user to install `pymatgen-analysis-defect` if this package is not in their environment.

Note: If `pymatgen-analysis-defect` is installed without `pybind11`, pip will raise errors.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ with open(path.join('dpgen', '_date.py'), 'w') as fp :
 install_requires=[
     'numpy>=1.14.3',
     'dpdata>=0.2.6',
-    'pymatgen>=2019.1.13',
+    'pymatgen>=2022.7.19',
     'ase',
     'monty>2.0.0',
     'paramiko',
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,8 @@ install_requires=[
     'dpdispatcher>=0.3.11',
     'netCDF4',
     'dargs>=0.2.9',
+    'pybind11>=2.4',
+    'pymatgen-analysis-defects'
 ]
 
 setuptools.setup(
--- a/tests/auto_test/test_interstitial.py
+++ b/tests/auto_test/test_interstitial.py
@@ -7,7 +7,13 @@ from monty.serialization import loadfn, dumpfn
 from pymatgen.core import Structure
 from pymatgen.io.vasp import Incar
 from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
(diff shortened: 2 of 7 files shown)


wanghan-iapcm:
In the test the verions of pymatgen is hard coded. Please update accordingly.
https://github.com/deepmodeling/dpgen/blob/8dea29ef125f66be9641afe5ac4970433a9c9ce1/.github/workflows/test.yml#L13

tests/auto_test/test_interstitial.py
```diff
@@ -7,7 +7,13 @@
...
 from pymatgen.io.vasp import Incar
 from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
-from pymatgen.analysis.defects.core import Interstitial as pmg_Interstitial
+try: 
+    from pymatgen.analysis.defects.core import Interstitial as pmg_Interstitial
+except ModuleNotFoundError:
+    print("pymatgen==2022.7.19 is the final release with `pymatgen.analysis.defects` module. Please check the version of pymatgen.")
+    print("Please install `pymatgen-analysis-defects`.")
+    print("Kindly reminder: `pybind11>=2.4` need to be installed previously.")
+    os._exit(0)
```
njzjz, line 16: Use raise instead

setup.py
```diff
@@ -29,6 +29,8 @@
     'dpdispatcher>=0.3.11',
     'netCDF4',
     'dargs>=0.2.9',
+    'pybind11>=2.4',
```
njzjz, line 32: Why do we need it?

HuangJiameng: Installing `pymatgen-analysis-defects` without `pybind11` will cause errors.

HuangJiameng:
Hmm... Some problems with GROMACS are raised. How can I fix it?

njzjz:
> Hmm... Some problems with GROMACS are raised. How can I fix it?

I don't think so.. It looks that some autotest modules are not compatible with the latest pymatgen.

njzjz:
I think we should fix it. There's two places that need attention.

kevinwenminion (approved):

HuangJiameng:
Could you fix the autotest modules to make it compatible with the latest pymatgen? Thanks. @kevinwenminion

ZLI-afk:
test

HuangJiameng:
Since fixes are made in #963, this pr should be closed.

PR #951, opened 2022-09-15
strict check simplify jdata

njzjz:
Since the simplify jdata is confirmed to be completed (if not, please remind me), we could enable the strict check.

### DIFF
--- a/dpgen/simplify/simplify.py
+++ b/dpgen/simplify/simplify.py
@@ -417,14 +417,6 @@ def run_iter(param_file, machine_file):
     jdata_arginfo = simplify_jdata_arginfo()
     jdata = normalize(jdata_arginfo, jdata)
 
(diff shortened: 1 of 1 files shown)


dpgen/simplify/simplify.py
```diff
@@ -413,6 +414,9 @@ def run_iter(param_file, machine_file):
         with open(machine_file, 'r') as fp:
             mdata = json.load(fp)
 
+    jdata_arginfo = simplify_jdata_arginfo()
+    jdata = normalize(jdata_arginfo, jdata)
+
     if jdata.get('pretty_print', False):
```
njzjz, line 420: ... What is `pretty_print` used for?

HuangJiameng (approved):

tuoping (approved):

Issue #957, opened 2022-09-18
[BUG] Generated the same coordinate file in make_fp step after turning on the merge_traj function 


wankiwi:
dpgen verison: 0.10.7.dev93+g72764d9.d20220915

After enabling the merge_traj function in model_deviation step, it will incorrectly generate the exact same coordinate file in the next make_fp step. All the coordinate files has the same hash, as below:
```
[02.fp] find . -name 'coord.xyz' | xargs md5sum 
517204ae3cd25bd7368b54f73abdac62  ./task.005.000000/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000001/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000002/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000003/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000004/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000005/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000006/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000007/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000008/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000009/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000010/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000011/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000012/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000013/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000014/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000015/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000016/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000017/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000018/coord.xyz
517204ae3cd25bd7368b54f73abdac62  ./task.005.000019/coord.xyz
......
517204ae3cd25bd7368b54f73abdac62  ./task.042.000134/coord.xyz
[...]

PR #958, opened 2022-09-18
Fix: correct confs for fp with merge_traj 

HuangJiameng:
Fix #957

PR #961, opened 2022-09-21
Supplementary unittests for merge_traj function to detect issue #957

Chengqian-Zhang:
The added unittest is to check that the process of extracting the configuration from model_devi.out to generate the first principles calculation task is correct.
When issue #957 exists:
j2vzixdu0g
When issue #957 is solved:
uzvvdHkD29


tests/generator/test_make_fp.py
```diff
@@ -3,13 +3,15 @@
 import numpy as np
 import unittest
 import importlib
+import textwrap
```
HuangJiameng, line 6: Why `textwrap` is imported?

Chengqian-Zhang: image
I use it here

Chengqian-Zhang:
> 

I think it is ok.

Chengqian-Zhang:
I think #958 can be replaced by this pr.

Chengqian-Zhang:
I think https://github.com/deepmodeling/dpgen/pull/958 can be replaced by this pr. @HuangJiameng

dpgen/generator/run.py
```diff
@@ -2190,8 +2190,11 @@ def _trust_limitation_check(sys_idx, lim):
             cwd = os.getcwd()
             os.chdir(fp_task_path)
             if cluster_cutoff is None:
-                if model_devi_engine in ("lammps", "gromacs"):
-                    os.symlink(os.path.relpath(conf_name), 'conf.dump')
+                if model_devi_engine in ("lammps", "gromacs"):                   
+                    if model_devi_merge_traj:
+                        conf_sys.to("lammps/lmp", "conf.dump")
```
njzjz, line 2195: I see line 2193 has `gromacs`, but I think this key only applies to `lammps`?

tests/generator/test_make_fp.py
njzjz: This method is used to remove whitespaces but I don't see any whitespace in line 304-315.

tests/generator/test_make_fp.py
```diff
@@ -231,7 +233,6 @@ def _write_lammps_dump(sys, dump_file, f_idx = 0) :
         for ii in range(natoms) :
             fp.write('%d %d %f %f %f\n' % (ii+1, atype[ii]+1, coord[ii][0], coord[ii][1], coord[ii][2]))
 
-
```
njzjz, line 234: This change should be reverted. Using 2 blank lines is the correct style.

dpgen/generator/run.py
HuangJiameng: Yes, it only applies to `lammps`. Since `gromacs` shares the traj format `lammps/dump`, see github.com/deepmodeling/dpgen/blob/devel/dpgen/generator/run.py#L2148-L2150, they are treated in the same way here. But `merge_traj` don't support `gromacs`, so I will divide them.

njzjz (approved):

HuangJiameng (approved):

AnguseZhang (approved):

PR #963, opened 2022-09-21
Fix incompatibility of autotest module with latest pymatgen>2022.7.19

ZLI-afk:
Fix incompatibility of Interstitial.py and Vacancy.py with pymatgen>2022.7.9
add pymatgen-analysis-defect and pybind11 in setup.py.
add reminder for user to install pymatgen-analysis-defect if this package is not in their environment

### DIFF
--- a/dpgen/auto_test/Interstitial.py
+++ b/dpgen/auto_test/Interstitial.py
@@ -2,6 +2,7 @@ import glob
 import json
 import os
 import re
+import numpy as np
 
 from monty.serialization import loadfn, dumpfn
 from pymatgen.analysis.defects.generators import InterstitialGenerator
@@ -147,9 +148,10 @@ class Interstitial(Property):
                     os.remove(insert_element_task)
 
                 for ii in self.insert_ele:
-                    vds = InterstitialGenerator(ss, ii)
+                    pre_vds = InterstitialGenerator()
+                    vds = pre_vds.generate(ss, {self.insert_ele[0]: [[0, 0.2, 0.5]]})
                     for jj in vds:
-                        temp = jj.generate_defect_structure(self.supercell)
+                        temp = jj.get_supercell_structure(sc_mat=np.eye(3) * self.supercell)
                         smallest_distance = list(set(temp.distance_matrix.ravel()))[1]
                         if 'conf_filters' in self.parameter and 'min_dist' in self.parameter['conf_filters']:
                             min_dist = self.parameter['conf_filters']['min_dist']
@@ -188,6 +190,9 @@ class Interstitial(Property):
 
 
                 if 'bcc_self' in self.parameter and self.parameter['bcc_self']:
+                    super_size = self.supercell[0] * self.supercell[1] * self.supercell[2]
+                    num_atom = super_size * 2
+                    chl = -num_atom - 2
                     os.chdir(path_to_work)
                     with open('POSCAR', 'r') as fin:
                         fin.readline()
@@ -210,7 +215,7 @@ class Interstitial(Property):
                     with open(insert_element_task, 'a+') as fout:
                         print(self.insert_ele[0], file=fout)
                     dumpfn(self.supercell, 'supercell.json')
-                    pos_line[-2] = '%.6f' % float(latt_param/4/super_latt_param) + ' ' + '%.6f' % float(latt_param/2/super_latt_param) + ' 0.000000 ' + self.insert_ele[0]
+                    pos_line[chl] = '%.6f' % float(latt_param/4/super_latt_param) + ' ' + '%.6f' % float(latt_param/2/super_latt_param) + ' 0.000000 ' + self.insert_ele[0]
                     with open('POSCAR', 'w+') as fout:
                         for ii in pos_line:
                                 print(ii, file=fout)
@@ -224,7 +229,7 @@ class Interstitial(Property):
                     with open(insert_element_task, 'a+') as fout:
                         print(self.insert_ele[0], file=fout)
                     dumpfn(self.supercell, 'supercell.json')
-                    pos_line[-2] = '%.6f' % float(latt_param/2/super_latt_param) + ' ' + '%.6f' % float(latt_param/2/super_latt_param) + ' 0.000000 ' + self.insert_ele[0]
+                    pos_line[chl] = '%.6f' % float(latt_param/2/super_latt_param) + ' ' + '%.6f' % float(latt_param/2/super_latt_param) + ' 0.000000 ' + self.insert_ele[0]
                     with open('POSCAR', 'w+') as fout:
                         for ii in pos_line:
                                 print(ii, file=fout)
@@ -238,7 +243,7 @@ class Interstitial(Property):
                     with open(insert_element_task, 'a+') as fout:
                         print(self.insert_ele[0], file=fout)
                     dumpfn(self.supercell, 'supercell.json')
-                    pos_line[-2] = '%.6f' % float(latt_param/4/super_latt_param) + ' ' + '%.6f' % float(latt_param/4/super_latt_param) + ' ' + '%.6f' % float(latt_param/4/super_latt_param) + ' ' + self.insert_ele[0]
+                    pos_line[chl] = '%.6f' % float(latt_param/4/super_latt_param) + ' ' + '%.6f' % float(latt_param/4/super_latt_param) + ' ' + '%.6f' % float(latt_param/4/super_latt_param) + ' ' + self.insert_ele[0]
                     with open('POSCAR', 'w+') as fout:
                         for ii in pos_line:
                                 print(ii, file=fout)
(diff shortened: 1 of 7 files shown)


dpgen/auto_test/Interstitial.py
```diff
@@ -2,9 +2,17 @@
 import json
 import os
 import re
+import numpy as np
 
 from monty.serialization import loadfn, dumpfn
 from pymatgen.analysis.defects.generators import InterstitialGenerator
+try:
+    from pymatgen.analysis.defects.generators import InterstitialGenerator
```
njzjz, line 10: Duplicate of line 8.

Also, I think we do not need to catch the error, as the dependency has already been added into `setup.py`.

tests/auto_test/test_vacancy.py
```diff
@@ -8,6 +8,13 @@
 from pymatgen.io.vasp import Incar
 from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
 from pymatgen.analysis.defects.core import Vacancy as pmg_Vacancy
+try:
+    from pymatgen.analysis.defects.core import Vacancy as pmg_Vacancy
```
njzjz, line 12: Duplicate of line 10.

njzjz (changes requested):
Read the docs should bump the Python version. See https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python

HuangJiameng:
The title should be 2022.7.19?

HuangJiameng (commented):
Should the title be 2022.7.19?

ZLI-afk:
> 

My mistake. already fixed, thanks

HuangJiameng:
@njzjz Do you know how to change the python version to >=3.8 when building readthedocs in checks?

njzjz:
> @njzjz Do you know how to change the python version to >=3.8 when building readthedocs in checks?

See https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python

HuangJiameng:
> > @njzjz Do you know how to change the python version to >=3.8 when building readthedocs in checks?
> 
> See https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python

Hmm...Where is `environment.yml` ?

njzjz:
> > > @njzjz Do you know how to change the python version to >=3.8 when building readthedocs in checks?
> > 
> > 
> > See https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python
> 
> Hmm...Where is `environment.yml` ?

If you mean `.readthedocs.yml`, you need to add it to the root directory. There is an example in that page.

njzjz (approved):

HuangJiameng:
@wanghan-iapcm Due to changes in pymatgen, we will not be able to support python versions below 3.8, which is against the [rules](https://docs.deepmodeling.com/projects/deepmd/en/master/development/coding-conventions.html#rules). But it seems that there is no better way. Please check.

HuangJiameng (approved):

setup.py
```diff
@@ -29,6 +29,8 @@
     'dpdispatcher>=0.3.11',
     'netCDF4',
     'dargs>=0.2.9',
+    'pybind11>=2.4',
```
njzjz, line 32: @HuangJiameng I didn't get errors when I tried to install `pymatgen-analysis-defects`. Please check again.

njzjz (changes requested):

wanghan-iapcm:
> @wanghan-iapcm Due to changes in pymatgen, we will not be able to support python versions below 3.8, which is against the [rules](https://docs.deepmodeling.com/projects/deepmd/en/master/development/coding-conventions.html#rules). But it seems that there is no better way. Please check it.

I agree to upgrade the python version requirement of this project to >= 3.8

dpgen/auto_test/Interstitial.py
```diff
@@ -147,9 +148,10 @@ def make_confs(self,
                     os.remove(insert_element_task)
 
                 for ii in self.insert_ele:
-                    vds = InterstitialGenerator(ss, ii)
+                    pre_vds = InterstitialGenerator()
+                    vds = pre_vds.generate(ss, {self.insert_ele[0]: [[0, 0.2, 0.5]]})
```
wanghan-iapcm, line 152: Only the first element of the `insert_ele` is inserted to the system. Its it true?  I do not how other elements in `insert_ele` is used in the new code.

dpgen/auto_test/Interstitial.py
```diff
@@ -147,9 +148,10 @@ def make_confs(self,
...
                     for jj in vds:
-                        temp = jj.generate_defect_structure(self.supercell)
+                        temp = jj.get_supercell_structure(sc_mat=np.eye(3) * self.supercell)
```
wanghan-iapcm, line 154: you may want to use `np.diag`...

dpgen/auto_test/Interstitial.py
```diff
@@ -210,7 +215,7 @@ def make_confs(self,
                     with open(insert_element_task, 'a+') as fout:
                         print(self.insert_ele[0], file=fout)
                     dumpfn(self.supercell, 'supercell.json')
-                    pos_line[-2] = '%.6f' % float(latt_param/4/super_latt_param) + ' ' + '%.6f' % float(latt_param/2/super_latt_param) + ' 0.000000 ' + self.insert_ele[0]
+                    pos_line[chl] = '%.6f' % float(latt_param/4/super_latt_param) + ' ' + '%.6f' % float(latt_param/2/super_latt_param) + ' 0.000000 ' + self.insert_ele[0]
```
wanghan-iapcm, line 218: Could you please explain this change? obviously chl != -2

dpgen/auto_test/Vacancy.py
```diff
@@ -134,10 +135,11 @@ def make_confs(self,
                 else:
                     ss = Structure.from_file(equi_contcar)
 
-                vds = VacancyGenerator(ss)
+                pre_vds = VacancyGenerator()
+                vds = pre_vds.generate(ss)
                 dss = []
                 for jj in vds:
-                    dss.append(jj.generate_defect_structure(self.supercell))
+                    dss.append(jj.get_supercell_structure(sc_mat=np.eye(3)*self.supercell))
```
wanghan-iapcm, line 142: You may want to use `np.diag`

dpgen/auto_test/Interstitial.py
wanghan-iapcm: [0, 0.2, 0.5] is hard coded. Could you please explain this parameter? In any case a user wants to change it?

ZLI-afk: It is the coordinates of the initial insertion. Since the insertions of the previous version of the Pymagen were not well defined, the function of this part of the codes was accomplished by random insertion of an atom into the bulk and then revising its position directly on the POSCAR. It was a compromise approach until the new Pymatgen come out where the position and type of the inserted atom can be specified easily. Thus, this PR just aims to make current codes compatible with the new Pymatgen, and we will rewrite this py later to make it more robust.

dpgen/auto_test/Interstitial.py
ZLI-afk: The variable chl is defined at line 195, which specifies the insertion atom line position in the POSCAR to be changed. This number is different from that of the previous version of Pymatgen.

dpgen/auto_test/Interstitial.py
ZLI-afk: Thanks, I didn't recall that.

dpgen/auto_test/Interstitial.py
wanghan-iapcm: It seems that the old behavior of interstitial generation is no longer supported in the latest pymatgen-analyze-defect, is it true?

ZLI-afk: YES, all defect-related functions in the Pymatgen after v2022.7.19 are transferred to the pymatgen-analysis-defects plugin. Additionally, this part was rewritten and some of the old functions and class methods are no longer valid.

HuangJiameng (changes requested):
Please check the `SyntaxError`.

dpgen/auto_test/Interstitial.py
wanghan-iapcm: Then what is the plan of pymatgen? Currently you provide the type `self.insert_ele[0]` and the coordinate `[[0, 0.2, 0.5]]` of the interstitial atom. Notice that the old behavior is to exhausitvely search all possible interstitial sites.

ZLI-afk: I have changed self.insert_ele[0] into ii in self.insert_ele so that it can iterate over the list. As for the coordinate [0, 0.2, 0.5], since the new version of pymatgen no longer provides the function to search interstitial sites automatically, I offer an initial input coordinate [0.1, 0.1, 0.1] which won't possibly conflict with any lattice atom to kick off the code.
If this plan is still not good enough, I will rewrite the current make_confs method to be better compatible with the new pymatgen

wanghan-iapcm: It would be better to directly ask the pymatgen people how to reproduce the old behavior with the new code.

Providing the coordinate like `[0.1, 0.1, 0.1] ` only gives you the  interstitial site nearest to `[0.1, 0.1, 0.1] ` , am I correct?

wanghan-iapcm: Thank you very much for the explanation. I accept the changes in this PR.

wanghan-iapcm (approved):

njzjz (approved):

HuangJiameng (approved):

AnguseZhang:
If we let DP-GEN not support python 3.7 anymore, we should update DP-GEN to 0.11.0

PR #989, opened 2022-10-09
fix: modify the reading of parameter 'fp_dpks_descriptor' in run.py

pxlxingliang:
In ABACUS-DEEPKS, ABACUS need the deepks descriptor file when one of 'deepks_out_labels' and 'deepks_scf' is set to 1.

### DIFF
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -2672,8 +2672,8 @@ def make_fp_abacus_scf(iter_index,
             if fp_params['basis_type'] == 'lcao':
                 assert('fp_orb_files' in jdata and type(jdata['fp_orb_files']) == list and len(jdata['fp_orb_files']) == len(fp_pp_files))
                 fp_orb_files = jdata['fp_orb_files']
-        dpks_out_labels = 0 if 'deepks_out_labels' not in fp_params else fp_params['deepks_out_labels']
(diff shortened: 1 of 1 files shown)


dpgen/generator/run.py
```diff
@@ -2672,10 +2672,11 @@ def make_fp_abacus_scf(iter_index,
             if fp_params['basis_type'] == 'lcao':
                 assert('fp_orb_files' in jdata and type(jdata['fp_orb_files']) == list and len(jdata['fp_orb_files']) == len(fp_pp_files))
                 fp_orb_files = jdata['fp_orb_files']
-        if 'deepks_out_labels' in fp_params:
-            if fp_params['deepks_out_labels'] == 1:
-                assert('fp_dpks_descriptor' in jdata and type(jdata['fp_dpks_descriptor']) == str)
-                fp_dpks_descriptor = jdata['fp_dpks_descriptor']
+        dpks_out_labels = 0 if 'deepks_out_labels' not in fp_params else fp_params['deepks_out_labels']
```
njzjz, line 2675: This line can be written as
```suggestion
        dpks_out_labels = fp_params.get('deepks_out_labels', 0)
```

Same for lines below.

njzjz (changes requested):

dpgen/generator/run.py
pxlxingliang: ok

njzjz (approved):

PR #994, opened 2022-10-12
simplify: support list for pick_data

njzjz:


HuangJiameng (approved):

wanghan-iapcm (commented):
I want to confirm that the convention used for `pick_data` is exactly the same as `init_data_sys `

njzjz:
Generally speaking, yes, but I remember that `init_data_sys` cannot be an absolute path.

PR #1055, opened 2022-11-29
add training_reuse_decay_steps

tuoping:
When starting with init-model, add a "training_reuse_decay_steps" parameter.


dpgen/generator/run.py
```diff
@@ -376,6 +377,7 @@ def make_train (iter_index,
         if jinput['loss'].get('start_pref_f') is not None:
             jinput['loss']['start_pref_f'] = training_reuse_start_pref_f
         jinput['learning_rate']['start_lr'] = training_reuse_start_lr
+        jinput['learning_rate']['decay_steps'] = training_reuse_decay_steps
```
njzjz, line 380: This is a breaking change

njzjz (changes requested):

PR #1059, opened 2022-12-01
Merge devel into master

amcadmus:


AnguseZhang (approved):
Agree with this PR. 

Meanwhile, this PR will introduce a breaking change to DP-GEN, since we will not support python3.7 anymore. So we should release 0.11.0, instead of 0.10.7, shall we?

njzjz (approved):

njzjz:
> Meanwhile, this PR will introduce a breaking change to DP-GEN, since we will not support python3.7 anymore.

This reminds me to update the documentation in #1060.

njzjz:
warning: #957 has not been fixed yet.

amcadmus:
> Agree with this PR.
> 
> Meanwhile, this PR will introduce a breaking change to DP-GEN, since we will not support python3.7 anymore. So we should release 0.11.0, instead of 0.10.7, shall we?

I agree. it will be released as 0.11.0

PR #1082, opened 2022-12-20
fix pass action

njzjz:
The original implementation does not work. For example, see #1080.


njzjz:
It fails, meaning it works.

PR #1083, opened 2022-12-21
"fp_task_min" key in param.json is not working.

tuoping:
"fp_task_min" key in param.json is not working. Because the "fp_task_min" value passed to _make_fp_vasp_inner function is wrong.  Fixed it.

Signed-off-by: tuoping <80671886+tuoping@users.noreply.github.com>


dpgen/generator/run.py
```diff
@@ -2574,12 +2574,7 @@ def _make_fp_vasp_configs(iter_index,
 
     modd_path = os.path.join(iter_name, model_devi_name)
     task_min = -1
-    if os.path.isfile(os.path.join(modd_path, 'cur_job.json')) :
-        cur_job = json.load(open(os.path.join(modd_path, 'cur_job.json'), 'r'))
-        if 'task_min' in cur_job :
-            task_min = cur_job['task_min']
-    else:
-        cur_job = {}
```
wanghan-iapcm, line 2582: why do you disable this part of code?

tuoping: Becuase I did not find a item "task_min" in my cur_job.json. Nor did I find the code writing an item "task_min" into the cur_job.json file.

wanghan-iapcm: The user may provide the key `task_min` in their job description.

wanghan-iapcm: BTW your PR is breaking UTs.

tuoping: Yes, this is puzzling, because I only changed this part. I think I should make the changes based on devel branch instead of master. I'm closing this.

PR #1085, opened 2022-12-22
fix compatibility with NumPy and pymatgen

njzjz:


wanghan-iapcm (approved):
I am surprised that the pymattgen has changed the interface....

PR #1091, opened 2022-12-29
fix doc: "structures" should be a list

HuangJiameng:
The type of parameter "structures" in `relaxation.json` of `dpgen autotest` should be a list. Fix the wrong type in Autotest Overview.

### DIFF
--- a/doc/autotest/Auto-test.md
+++ b/doc/autotest/Auto-test.md
@@ -15,7 +15,7 @@ If, for some reasons, the main program terminated at stage `run`, one can easily
 `relax.json` is the parameter file. An example for `deepmd` relaxation is given as:
 ```json
 {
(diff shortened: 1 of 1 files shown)


AnguseZhang:
Do we also need to optimize the error log if users provide wrong types of inputs? Probably use `dargs`? @HuangJiameng

njzjz (approved):

njzjz:
> Do we also need to optimize the error log if users provide wrong types of inputs?

You can open a new issue to track it.

HuangJiameng:
> Do we also need to optimize the error log if users provide wrong types of inputs? Probably use `dargs`? @HuangJiameng

The check on input argument dict hasn't been applied to `autotest` yet. An issue is welcomed.

PR #1092, opened 2022-12-29
Support forward files for simplify

HuangJiameng:
Support "user_forward_files" for `simplify` in the `02.fp` step. Use abspath to create symlinks to avoid unexpected problems.

### DIFF
--- a/dpgen/simplify/simplify.py
+++ b/dpgen/simplify/simplify.py
@@ -411,6 +411,10 @@ def make_fp(iter_index, jdata, mdata):
     else:
         make_fp_configs(iter_index, jdata)
         make_fp_calculation(iter_index, jdata)
+    # Copy user defined forward_files
+    iter_name = make_iter_name(iter_index)
+    work_path = os.path.join(iter_name, fp_name)
+    symlink_user_forward_files(mdata=mdata, task_type="fp", work_path=work_path)
 
 
 def run_iter(param_file, machine_file):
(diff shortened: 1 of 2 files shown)


AnguseZhang (approved):

dpgen/simplify/simplify.py
```diff
@@ -411,6 +411,10 @@ def make_fp(iter_index, jdata, mdata):
     else:
         make_fp_configs(iter_index, jdata)
         make_fp_calculation(iter_index, jdata)
+    # Copy user defined forward_files
+    iter_name = make_iter_name(iter_index)
+    work_path = os.path.join(iter_name, fp_name)
+    symlink_user_forward_files(mdata=mdata, task_type="fp", work_path=work_path)
```
njzjz, line 417: These lines should be executed only when `labeled` is `False`.

njzjz: Also, do we already have corresponding parameters?

HuangJiameng: > Also, do we already have corresponding parameters?

It can be set in `machine.json`, just like `run`—tested work.

njzjz (approved):

ysu-007:
Have you tested it? It is still not working for me.

HuangJiameng:
> Have you tested it? It is still not working for me.

img_v2_9b34c63e-779a-466d-92d8-4d22937a049g

ysu-007:
Are the parameters in the machine.json right? 
image

HuangJiameng:
> Are the parameters in the machine.json right? image

The correct parameter is "user_forward_files".

ysu-007:
> > Are the parameters in the machine.json right? image
> 
> The correct parameter is "user_forward_files".

It works now, thanks

PR #1105, opened 2023-01-05
Replace "python3" in command by sys.executable

HuangJiameng:
fix #1099

### DIFF
--- a/dpgen/data/gen.py
+++ b/dpgen/data/gen.py
@@ -708,6 +708,7 @@ def pert_scaled(jdata) :
         fp_style = "abacus"
         poscar_name = "STRU"
     pert_cmd = 'python3 ' + pert_cmd + ' -etmax %f -ofmt %s %s %d %f > /dev/null' %(pert_box, fp_style, poscar_name, pert_numb, pert_atom)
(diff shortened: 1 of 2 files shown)


dpgen/data/gen.py
```diff
@@ -708,6 +708,7 @@ def pert_scaled(jdata) :
         fp_style = "abacus"
         poscar_name = "STRU"
     pert_cmd = 'python3 ' + pert_cmd + ' -etmax %f -ofmt %s %s %d %f > /dev/null' %(pert_box, fp_style, poscar_name, pert_numb, pert_atom)
+    pert_cmd = pert_cmd.replace('python3', sys.executable)
```
njzjz, line 711: This will also replace other part of the command. For example, the command in #1099 is `python3 /public/home/sugon/.local/lib/python3.9/site-packages/dpgen/data/tools/create_random_disturb.py -etmax 0.030000 -ofmt vasp POSCAR 30 0.010000 > /dev/null` which contains two `python3`. It should be more direct.

njzjz (changes requested):

njzjz (approved):

PR #1111, opened 2023-01-08
fix restart trajectory for `merge_traj`

wankiwi:
When turning on the merge_traj function, if the procedure is interrupted in the run_model_devi stage, the old trajectory file will be overwritten after it is rerun. The incomplete trajectory file will cause the following error in the next post_model_devi stage:
`.../site-packages/dpdata/lammps/dump.py:176: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
system['coords'] = np.array(system['coords'])
DeepModeling
`
This PR repaired this question.


njzjz:
@HuangJiameng could you review this PR?

HuangJiameng (approved):

HuangJiameng (approved):

PR #1124, opened 2023-01-25
add and run pre-commit to format codes

njzjz:
- add pre-commit config, including black and common fixes.
- run pre-commit.
- manually fix a YAML file.
- skip check json and broken symlinks.


wanghan-iapcm (approved):
I add Yuzhi as reviewer

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    hooks:
+    # there are many log files in tests
+    # TODO: seperate py files and log files
+    -   id: trailing-whitespace
+        exclude: "^tests/.*$"
+    -   id: end-of-file-fixer
+        exclude: "^tests/.*$"
+    -   id: check-yaml
+        exclude: "^conda/.*$"
+    #-   id: check-json
```
AnguseZhang, line 15: Why is `check-json` skipped?

AnguseZhang (approved):

.pre-commit-config.yaml
njzjz: `check-json` does not allow duplicated keys. See https://github.com/pre-commit/pre-commit-hooks/blob/5191f112dfd133e97acba8c2162c37d8613cf54b/pre_commit_hooks/check_json.py#L15

njzjz: If we enable this hook, we need to rename _comment to _comment1, _comment2, ..., _commentN. We don't have the time to do so, but this line can be uncommented in a separated PR.

PR #1133, opened 2023-02-02
add isort to sort imports

njzjz:
This is an automatic tool to format and sort imports. See https://github.com/PyCQA/isort for details.


njzjz:
> dargs.dargs.ArgumentKeyError: [at location `train/machine/remote_profile`] undefined key `email` is not allowed in strict mode

@ACoderHIT Has this key been deprecated?

PR #1137, opened 2023-02-10
Updated the `revise_lmp_input_dump` function.

panxiang126:
Added the `model_devi_merge_traj` parameter to allow the merging of trajectories into one file using the lmp_input template.


panxiang126 (commented):
Add default option to `model_devi_merge_traj`: `False`

njzjz (approved):

PR #1149, opened 2023-03-09
Lammps template for MD+MC

wujing81:
Lammps template for MD+MC


Vibsteamer (changes requested):
The attatched param.json together with the template now dump-and-check in one loop_step:

1) 50 MC snapshots from the last MD_snapshot in the last loop, **many of them are repeated images** as being not accepted by Metroplis criterion.
2) then, 100 MD snapshots from the last accepted MC_snapshot.

the **sampling weights**  (then candidates and labelds) on physical structures could be correspondingly **biased**,

Please make sure this is of your expectation.

PR #1153, opened 2023-03-16
fix(abacus): do not write to deepmd raw/npy, if no abacus data in post_fp_abacus_scf()

pxlxingliang:


wanghan-iapcm (commented):
Could you please provide a UT for the case if no abacus data is generated

pxlxingliang:
> Could you please provide a UT for the case if no abacus data is generated

Ok, the unit test is added.

PR #1176, opened 2023-04-06
seems log format in abacus have changed

floatingCatty:
now the steps is recorded as 
"-------------------------------------------
 STEP OF RELAXATION : 1
 -------------------------------------------"
so this might requires modification.


pxlxingliang:
Hi, @floatingCatty, this bug has been fixed in #1169, and please pull the latest devel branch.

PR #1209, opened 2023-05-06
support automatic training_reuse_old_ratio

njzjz:
set the default value to `auto` instead of raising errors


njzjz:
@robinzyb it seems that unit test for cp2k is broken after dpdata v0.2.15 is released.

Vibsteamer (approved):

PR #1212, opened 2023-05-08
fix bug in make_supercell_abacus

Liu-RX:
The `from_struct["atom_types"]` used to be a list but was changed to numpy array lately, thereby introducing a bug in `make_supercell_abacus` function.
Also, the key `types` is changed to `atom_types` outside the function, but not changed in the function.
 This commit fix the two bugs.


dpgen/generator/lib/abacus_scf.py
```diff
@@ -392,31 +393,49 @@ def get_abacus_STRU(STRU, INPUT=None, n_ele=None):
...
                         + from_struct["cells"][0] * ix
                         + from_struct["cells"][1] * iy
                         + from_struct["cells"][2] * iz
                     )
-                    from_struct["coords"] = np.vstack([from_struct["coords"], coord])
-    from_struct["atom_numbs"] = [
+                    if ia == 0 and ix == 0 and iy == 0 and iz == 0:
+                        new_coord = np.array(coord)
+                    else:
+                        new_coord = np.vstack([new_coord, coord])
```
njzjz, line 428: `vstack` is expensive. It's recommended to call `vstack` only once.

Liu-RX: I have replaced the `vstack` with value assignment in a pre-allocated zero array.

njzjz (approved):

PR #1241, opened 2023-06-09
fix: description in README.md

cherushui:


njzjz (approved):

njzjz:
Thanks.

`doc/overview/overview.md` has the same problem. CC @HuangJiameng

PR #1287, opened 2023-07-25
Using dict.get methods to get machine.local_root

dingzhaohan:
If local_root isn't `./`, dpgen will raise error and exit, it seems local_root must be './', so I try to get a default value if machine.json doesn't provided it.


wanghan-iapcm:
The dispatcher module is not maintained any more. One is supposed to use https://github.com/deepmodeling/dpdispatcher. 
@njzjz shall we remove dpgen/dpdispatcher?

dingzhaohan:
func make_submission(dpgen/dispatcher/Dispatcher.py) is called in dpgen/generator/run.py. In make_submission, it first checks local_root, then initialize dpdispatcher.

njzjz:
> The dispatcher module is not maintained any more. One is supposed to use https://github.com/deepmodeling/dpdispatcher. 
> 
> @njzjz shall we remove dpgen/dpdispatcher? 

It now only contains a wrapper of dpdispatcher.

njzjz (approved):

PR #1302, opened 2023-08-04
Update {Elastic,run,surf,Vasp,vasp}.py: To be compatible with Pymatgen

ZhouXY-PKU:
To be compatible with old and new versions of Pymatgen  (https://github.com/materialsproject/pymatgen/pull/3158)


njzjz (changes requested):
This PR breaks compatibility with old versions, so the minimal version in requirements should be bumped.

ZhouXY-PKU:
> This PR breaks compatibility with old versions, so the minimal version in requirements should be bumped.

Thanks for your review! I think the minimal version in requirements is already bumped now.

wanghan-iapcm:
Please fix the UTs. Thanks

ZhouXY-PKU:
> Please fix the UTs. Thanks

Sorry that there have been some mistakes. I think it can work now.

> Please fix the UTs. Thanks

Sorry that there have been some mistakes. I think it can work now.

dpgen/auto_test/Elastic.py
```diff
@@ -221,7 +221,12 @@ def post_process(self, task_list):
                 kspacing = incar.get("KSPACING")
                 kgamma = incar.get("KGAMMA", False)
                 ret = vasp.make_kspacing_kpoints(poscar_start, kspacing, kgamma)
-                kp = Kpoints.from_string(ret)
+                pymgv = "old"
```
wanghan-iapcm, line 224: I would not recommend introducing the pymgv variable, but rather use `try`...`except` everywhere various interfaces are introduced in different pymatgen versions.  suppose pymatgen introduce a third variety of `from_str`, we need to add a new value other than the "old" "new" to pymgv, "new2"? which is anyway not a nice name. 

Another option is that we provide an interface for calling with different versions of pymatgen, then the `try`...`except` only appear once in the code.

ZhouXY-PKU: Thanks for your recommendation! The extra variable `pymgv` is fully replaced by  `try`...`except` now. (I have tried to get the version of Pymatgen directly, but it seems that Pymatgen had no member like `__version__` at all.)

njzjz (approved):

PR #1322, opened 2023-09-01
dprc: add `restart_from_iter` option

njzjz:
Add the `restart_from_iter` option in each iteration of DPRc simulations to restart from a previous iteration instead of initial structures.


wanghan-iapcm (commented):
we may want to explicitly say in the doc that the `restart_from_iter` only support dprc with amber.

njzjz:
This argument is inside the amber variant, so the path should be clear.

BTW, I found the documentation is down...

njzjz:
> BTW, I found the documentation is down...

It seems that the error is caused by https://github.com/sphinx-doc/sphinx/issues/11662

wanghan-iapcm (approved):

PR #1340, opened 2023-09-21
support training with -f or -t argument

### LINKED ISSUE #1122: Finetune pretrained DPA-1 model with dpgen
### Discussed in https://github.com/deepmodeling/dpgen/discussions/1116

<div type='discussions-op-text'>

<sup>Originally posted by **KeweiZhao** January 16, 2023</sup>
Hello,

I'm wondering if there is a way to finetune a pretrained DPA-1 model with dpgen (e.g. OC_10M.pb). I haven't find a way to pass the "--finetune" argument to dp using dpgen.

Thanks,
Kewei</div>

njzjz:
Fix #1122.


njzjz:
@wanghan-iapcm Discussion is required: whether these flags should only work in iteration 0 so that they can work with `training_reuse_iter` without modifying input files.

wanghan-iapcm:
> @wanghan-iapcm Discussion is required: whether these flags should only work in iteration 0 so that they can work with `training_reuse_iter` without modifying input files.

Yes, they should only work for iteration 0. Actually the finetune does two things: 1. statistic the training data and set the new energy bias in the model, 2. train the new model with the `--init-model` mode. 

You may want to check the implementation in dpgen2 https://github.com/deepmodeling/dpgen2/pull/152

wanghan-iapcm (approved):

PR #1349, opened 2023-10-08
Add `dpgui` entry point and `dpgen gui` CLI

njzjz:
See https://github.com/deepmodeling/dpgui/issues/270 for details.


njzjz:
There seems to be an issue here. Python will dump NaN as `{"type": NaN}`. However, JavaScript doesn't think it a valid JSON string.

Python:
```py
>>> import json
>>> json.dumps({"type":float("nan")})
'{"type": NaN}'
```

JavaScript:
```js
> JSON.parse('{"type": NaN}')
VM1054:1 Uncaught SyntaxError: Unexpected token 'N', "{"type": NaN}" is not valid JSON
    at JSON.parse (<anonymous>)
    at <anonymous>:1:6
```

njzjz:
> Python will dump NaN as `{"type": NaN}`.

Finally, I changed `NaN` to `1e10` to avoid this issue.

PR #1372, opened 2023-10-31
docs: update installation documentation

njzjz:
Since the dpgen conda package on the deepmodeling channel and the offline packages are not maintained anymore, this PR removes the related documentation.


njzjz:
Btw, we may release a new version. The last version was half a year ago.

wanghan-iapcm (approved):

wanghan-iapcm:
> Btw, we may release a new version. The last version was half a year ago.

I agree

PR #1375, opened 2023-10-31
move to ruff formatter

njzjz:
See https://astral.sh/blog/the-ruff-formatter.


wanghan-iapcm (approved):
It actually looks nicer than black

Issue #1246, opened 2023-06-16
[Feature Request] use MultiSystems in post_fp_vasp function


erichuangyf:
**Summary**
Add MultiSystems support while posting vasp jobs

**Detailed Description**
I am currently investigating the feasibility of incorporating multiple vacancy structures with different formulas into a single sys_configs tag. However, I encountered error message #306 due to inconsistent atom numbers in this mixed system setup. This behavior was expected, but I have identified a potential solution to address it.

To resolve the issue, I suggest modifying the post_fp_vasp function in dpgen/dpgen/generator/run.py:3914 as follows: replace all_sys = None with all_sys = dpdata.MultiSystems(type_map=jdata["type_map"]). This adjustment utilizes the MultiSystems functionality from the dpdata library, effectively preventing the single-labeled system complaining about formula difference.

While this change successfully resolves the problem, I have concerns regarding potential risks associated with introducing multiple systems in one fp task. Therefore, I would like to inquire about any potential harm or drawbacks in this approach. If there are no significant concerns, I kindly request your consideration in applying this code change, enabling the mixing of vacancy structures into one training system.

njzjz:
This seems reasonable to me. But we need to support `MultiSystems` for `dpgen collect` before this change.

njzjz:
> But we need to support `MultiSystems` for `dpgen collect` before this change.

Done in #1422.

`MultiSystems` can be introduced then.

PR #1513, opened 2024-04-10
Modify the post_fp_cp2k function to generate multisystem data. Allows to run different system at one single iteration model_devi step.

dulinhan:
Modify the post_fp_cp2k function to generate multisystem data. Allows to run different system at one single iteration model_devi step.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,8 @@ classifiers = [
 ]
 dependencies = [
     'numpy>=1.14.3',
-    'dpdata>=0.2.16',
+    'dpdata>=0.2.17',
+    'cp2kdata>=0.6.6',
     'pymatgen>=2022.11.1',
     'ase',
     'monty>2.0.0',
--- a/tests/generator/test_post_fp.py
+++ b/tests/generator/test_post_fp.py
@@ -281,7 +281,7 @@ class TestPostCP2K(unittest.TestCase, CompLabeledSys):
         post_fp(0, jdata)
         self.system_1 = dpdata.LabeledSystem("iter.000000/orig", fmt="deepmd/raw")
         self.system_2 = dpdata.LabeledSystem(
-            "iter.000000/02.fp/data.000", fmt="deepmd/raw"
+            "iter.000000/02.fp/data.000/C2H2N2", fmt="deepmd/raw"
         )
 
 
--- a/dpgen/generator/run.py
+++ b/dpgen/generator/run.py
@@ -4439,14 +4439,14 @@ def post_fp_cp2k(iter_index, jdata, rfailed=None):
         tcount += len(sys_output)
         all_sys = dpdata.MultiSystems(type_map=jdata["type_map"])
         for oo in sys_output:
-            _sys = dpdata.LabeledSystem(oo, fmt="cp2k/output")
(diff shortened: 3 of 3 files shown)


Yi-FanLi (commented):
Please delete the code you want to delete instead of commenting it and leaving it here. : )

Yi-FanLi:
Now it looks good!

Yi-FanLi (approved):

pyproject.toml
```diff
@@ -19,7 +19,8 @@ classifiers = [
 ]
 dependencies = [
     'numpy>=1.14.3',
-    'dpdata>=0.2.16',
+    'dpdata>=0.2.17',
```
njzjz, line 22: What's the reason to bump dpdata version?

dulinhan: To better support cp2kdata with the newer dpdata version.

robinzyb: it is ok to bump version here. I think any version of dpdata can support cp2kdata as long as the plugin function is implemented.

tests/generator/test_post_fp.py
```diff
@@ -281,7 +281,7 @@ def setUp(self):
         post_fp(0, jdata)
         self.system_1 = dpdata.LabeledSystem("iter.000000/orig", fmt="deepmd/raw")
         self.system_2 = dpdata.LabeledSystem(
-            "iter.000000/02.fp/data.000", fmt="deepmd/raw"
+            "iter.000000/02.fp/data.000/C2H2N2", fmt="deepmd/raw"
```
robinzyb, line 284: Is this ok to write multiple sys under same `sys_idx`? I remember that systems should have different suffixes, e.g. `000`, `001`

dulinhan: According to my test, the modified post_fp_cp2k function will create a folder like data.xxx/sys1, data.xxx/sys2 ... in iteration xxx. Fortunately, make_train function could identify several subfolders in data.xxx.

dulinhan: I think it is an expected behavior. I originally planned to modify 'make_train' in 'run.py', but it is compatible with multi-systems. Also keeping the original 'make_train' will not destroy the route with different fp methods, i.e. 'vasp' or 'QE'.

dulinhan: I think it is an expected behavior. I originally planned to modify `make_train` in `run.py`, but it is compatible with multi-systems. Also keeping the original `make_train` will not destroy the route with different fp methods, i.e. `vasp` or `QE`.

njzjz: xref: #1246

pyproject.toml
njzjz: The only concern is that cp2kdata still needs to be added to conda-forge. @robinzyb After I submit it, would you like to maintain it?

robinzyb: yeah. I can maintain it. Maybe I can get strated from submitting it to conda-forge. I will do this tomorrow

robinzyb (approved):

tests/generator/test_post_fp.py
robinzyb: good, I got no problem

njzjz (approved):

Issue #1462, opened 2024-01-27
[Feature Request] Support different backends for DeePMD-kit


njzjz:
### Summary

Support PyTorch backend for DeePMD-kit.

### Detailed Description

Although I have submitted https://github.com/deepmodeling/deepmd-kit/pull/3191 to make most of the things the same, the suffixes of the model and checkpoint files are different by design. These suffixes are hardcoded in the code.

https://github.com/deepmodeling/dpgen/blob/73926506688ac8f8939f065056e83ac51fb2601f/dpgen/generator/run.py#L804-L812
https://github.com/deepmodeling/dpgen/blob/73926506688ac8f8939f065056e83ac51fb2601f/dpgen/generator/run.py#L884-L889
https://github.com/deepmodeling/dpgen/blob/73926506688ac8f8939f065056e83ac51fb2601f/dpgen/generator/run.py#L1938
https://github.com/deepmodeling/dpgen/blob/73926506688ac8f8939f065056e83ac51fb2601f/dpgen/simplify/simplify.py#L188

Besides, glob doesn't work for `backward_files` - in dpdispatcher, some contexts like Bohrium don't support it.
So, may we consider adding a parameter to switch?

njzjz:
This is blocked by deepmodeling/deepmd-kit#3475.

thangckt:
hi @njzjz 

Did you solved this problem?
I made some changes to have option for choosing between backends TF and PT, in [this file](https://github.com/thangckt/dpgen/commit/28b8a49d2246070aedc563a1a3109ff42134a9b8)

If you find it helpful, I will make a PR. Otherwise, just omit

PR #1541, opened 2024-05-07
add option to select backends TF/PT

thangckt:
add a new key in `param.json` file
```
"train_backend": "pytorch"/"tensorflow",
```
relate to this issue #1462

## Summary by CodeRabbit

- **New Features**
	- Enhanced model training and development processes.
	- Introduced new functionalities for model deviation analysis with various system configurations.


dpgen/generator/run.py
```diff
@@ -125,6 +125,16 @@
 run_opt_file = os.path.join(ROOT_PATH, "generator/lib/calypso_run_opt.py")
 
 
+def _get_model_suffix(jdata) -> str:
+    """Return the model suffix based on the backend"""
+    backend = jdata.get("train_backend", "tensorflow")
+    if backend == "tensorflow":
+        suffix = ".pb"
+    elif backend == "pytorch":
+        suffix = ".pth"
```
njzjz, line 134: should use `else` to handle other cases.

dpgen/generator/run.py
```diff
@@ -125,6 +125,16 @@
 run_opt_file = os.path.join(ROOT_PATH, "generator/lib/calypso_run_opt.py")
 
 
+def _get_model_suffix(jdata) -> str:
+    """Return the model suffix based on the backend"""
+    backend = jdata.get("train_backend", "tensorflow")
```
njzjz, line 130: This argument should be added into `arginfo.py` for documentation and check.

dpgen/generator/run.py
```diff
@@ -799,17 +814,23 @@ def run_train(iter_index, jdata, mdata):
...
-    if jdata.get("dp_compress", False):
-        backward_files.append("frozen_model_compressed.pb")
+    if suffix == ".pb":
+        backward_files += [
+            "model.ckpt.meta",
+            "model.ckpt.index",
+            "model.ckpt.data-00000-of-00001",
+        ]
+        if jdata.get("dp_compress", False):
+            backward_files.append("frozen_model_compressed%s" % suffix)
```
njzjz, line 832: should handle the situation for pt.

For pt, the checkpoint file is `model.pt`; the compressed model file will be `frozen_model_compressed.pth`

dpgen/generator/run.py
```diff
@@ -762,9 +777,9 @@ def run_train(iter_index, jdata, mdata):
         if training_init_model:
             init_flag = " --init-model old/model.ckpt"
         elif training_init_frozen_model is not None:
-            init_flag = " --init-frz-model old/init.pb"
+            init_flag = " --init-frz-model old/init%s" % suffix
         elif training_finetune_model is not None:
-            init_flag = " --finetune old/init.pb"
+            init_flag = " --finetune old/init%s" % suffix
         command = f"{train_command} train {train_input_file}{extra_flags}"
```
njzjz, line 783: For pt, we use `dp --pt train`

dpgen/generator/run.py
```diff
@@ -880,13 +901,15 @@ def post_train(iter_index, jdata, mdata):
...
         return
     # symlink models
+    suffix = _get_model_suffix(jdata)
     for ii in range(numb_models):
-        if not jdata.get("dp_compress", False):
-            model_name = "frozen_model.pb"
-        else:
-            model_name = "frozen_model_compressed.pb"
+        model_name = "frozen_model%s" % suffix
+        if suffix == ".pb":
```
njzjz, line 907: should preserve the situation for pt. I think tf and pt share the same logic here.

njzjz (changes requested):

dpgen/generator/run.py
thangckt: I think this should be set from the key: 
```json
"train": [
        {
	        "command": "dp --pt"
        }
]
```

dpgen/generator/run.py
njzjz: It should be `model.ckpt.pt`

dpgen/generator/run.py
njzjz: This is okay, but needs documentation.

dpgen/generator/run.py
```diff
@@ -3770,6 +3838,8 @@ def make_fp_calculation(iter_index, jdata, mdata):
         make_fp_amber_diff(iter_index, jdata)
     elif fp_style == "custom":
         make_fp_custom(iter_index, jdata)
+    elif fp_style == "gpaw":
```
njzjz, line 3841: Could you submit gpaw in a different PR?

PR #1588, opened 2024-07-15
fix: uninitialized return values

wanghan-iapcm:


wanghan-iapcm:
close. i did not notice #1587

PR #1627, opened 2024-08-27
solve Path error on Windows

thangckt:
## Summary by CodeRabbit

- **Chores**
	- Improved code readability and maintainability through cleanup and simplification of functions.
	- Streamlined file path construction using the modern `Path` API.
	- Removed unnecessary operations related to file handling and training resources.


dpgen/generator/run.py
```diff
@@ -205,8 +205,8 @@ def copy_model(numb_model, prv_iter_index, cur_iter_index, suffix=".pb"):
             "graph.%03d%s" % (ii, suffix),
         )
         os.chdir(cwd)
-    with open(os.path.join(cur_train_path, "copied"), "w") as fp:
-        None
+    # with open(os.path.join(cur_train_path, "copied"), "w") as fp:
+    #     None
```
njzjz, line 209: Is there a reason to comment it?

thangckt: I dont know that command for what? It does nothing?

njzjz: It creates an empty file named `copied`, which is equivalent to `Path(os.path.join(cur_train_path, "copied")).touch()`.

I think this empty file is only to mark this step is skipped and the models are copied from the last iteration.

thangckt: ok

PR #1691, opened 2024-12-19
fix: bump the upper bound of deepmd version

### LINKED ISSUE #1688: [BUG] The process of `deepmd_version` needs to be updated
### Bug summary

Several codes like below needs to be updated.
https://github.com/deepmodeling/dpgen/blob/e02e6a2e8ee383955ddfa661bca2a716b2016add/dpgen/generator/run.py#L475-L477

https://github.com/deepmodeling/dpgen/blob/e02e6a2e8ee383955ddfa661bca2a716b2016add/dpgen/generator/run.py#L498-L501

### DP-GEN Version

devel

### Platform, Python Version, Remote Platform, etc

_No response_

### Input Files, Running Commands, Error Log, etc.

N/A

### Steps to Reproduce

N/A

### Further Information, Files, and Links

N/A

njzjz:
Fix #1688. Though I am not sure whether we should have an upper version.

## Summary by CodeRabbit

- **New Features**
	- Enhanced support for `deepmd` library versions 1.x to 3.x.
	- Improved handling of model deviation tasks with refined logic for training conditions.
	- Updated file management for better input/output file linking across tasks.

- **Bug Fixes**
	- Clarified error messages for unsupported library versions.

- **Documentation**
	- Updated documentation strings for argument compatibility and function behavior.

- **Refactor**
	- Adjusted function signatures to accommodate new parameters and improved logic.

### DIFF
--- a/dpgen/generator/arginfo.py
+++ b/dpgen/generator/arginfo.py
@@ -297,8 +297,8 @@ def model_devi_jobs_args() -> list[Argument]:
     doc_taup = "Coupling time of barostat (ps)."
     doc_model_devi_f_trust_lo = "Lower bound of forces for the selection. If dict, should be set for each index in sys_idx, respectively."
     doc_model_devi_f_trust_hi = "Upper bound of forces for the selection. If dict, should be set for each index in sys_idx, respectively."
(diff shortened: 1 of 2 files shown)


wanghan-iapcm (approved):
I do not think we need an upper bound for the version.

Issue #1888, opened 2026-06-21
run_mdata_arginfo rejects list-form machine sections accepted by runtime convert_mdata


SchrodingersCattt:
## Summary

`dpgen.generator.arginfo.run_mdata_arginfo()` rejects list-form `machine.json` sections for `train`, `model_devi`, and `fp`, but the actual DP-GEN runtime path accepts these sections as either dicts or lists/tuples.

This makes external validators/tools that rely on `run_mdata_arginfo()` stricter than `dpgen run` itself and can falsely reject machine files that DP-GEN runtime accepts.

## Observed inconsistency

In `dpgen==0.13.3`, `run_mdata_arginfo()` is defined as:

```python
def run_mdata_arginfo() -> Argument:
    return general_mdata_arginfo("run_mdata", ("train", "model_devi", "fp"))
```

and `general_mdata_arginfo()` creates these sections as `dict`:

```python
for task in tasks:
    sub_fields.append(
        Argument(
            task,
            dict,
            optional=False,
            sub_fields=mdata_arginfo(),
            doc=doc_mdata % task,
        )
    )
```

So dargs rejects:

```json
{
  "train": [
    {
      "command": "dp",
      "machine": {"batch_type": "Shell", "context_type": "ssh", "local_root": "./", "remote_root": "/path/to/work"},
      "resources": {"number_node": 1, "cpu_per_node": 4, "gpu_per_node": 1, "group_size": 1}
    }
  ],
  "model_devi": [
    {
      "command": "lmp",
      "machine": {"batch_type": "Shell", "context_type": "ssh", "local_root": "./", "remote_root": "/path/to/work"},
      "resources": {"number_node": 1, "cpu_per_node": 4, "gpu_per_node": 0, "group_size": 1}
    }
  ],
  "fp": [
    {
      "command": "vasp_std",
      "machine": {"batch_type": "Shell", "context_type": "ssh", "local_root": "./", "remote_root": "/path/to/work"},
      "resources": {"number_node": 1, "cpu_per_node": 32, "gpu_per_node": 0, "group_size": 1}
    }
  ]
}
```

with:

```text
[at root location] key `train` gets wrong value type, requires <dict> but list is not a dict
```

However, the actual runtime path in `dpgen/generator/run.py` does not validate `machine.json` with `run_mdata_arginfo()`. It loads machine data and then calls `convert_mdata(mdata)`:

```python
jdata = load_file(param_file)
mdata = load_file(machine_file)

[...]

PR #1889, opened 2026-06-21
fix: align mdata arginfo with runtime machine sections

SchrodingersCattt:
## Summary

- Align `general_mdata_arginfo()` with `convert_mdata()` runtime semantics for machine task sections.
- Allow `train`, `model_devi`, and `fp` to be either a dict or a list of dicts.
- Add regression tests covering dict-form, list-form, scalar rejection, and list-with-non-dict rejection.

## Related

Related to deepmodeling/dpgen#1888.

## Why

`dpgen run` accepts list-form machine sections through `convert_mdata()`:

```python
elif isinstance(mdata[task_type], (list, tuple)):
    task_data = mdata[task_type][0]
```

But `run_mdata_arginfo()` previously exposed these sections as dict-only, causing external validators/tools to falsely reject machine files that runtime accepts.

## Validation

- `python -m pytest tests/test_mdata_arginfo.py tests/test_check_examples.py tests/tools/test_convert_mdata.py -q`
- `python -m ruff check dpgen/arginfo.py tests/test_mdata_arginfo.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime validation for metadata task configurations, accepting either task sections as `dict` objects or as non-empty lists of `dict` objects.
  * Added clearer error messaging when metadata task inputs don’t match the expected runtime shape.

* **Tests**
  * Added new unit tests covering acceptance of valid task section formats.
  * Added coverage for rejection cases including lists with non-dict items, empty lists, and scalar inputs.

### DIFF
--- a/dpgen/arginfo.py
+++ b/dpgen/arginfo.py
@@ -1,5 +1,15 @@
 from dargs import Argument
 
+
+def _is_mdata_task_config(value):
+    """Return whether a machine task section matches runtime accepted shapes."""
+    if isinstance(value, dict):
+        return True
+    if isinstance(value, list):
+        return all(isinstance(item, dict) for item in value)
+    return False
+
+
 from dpgen.dispatcher.Dispatcher import mdata_arginfo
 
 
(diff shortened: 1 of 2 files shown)


njzjz (changes requested):
The list form has been deprecated and is just kept for compatibility. I don't think it is necessary to tell external tools that a deprecated usage is supported.

SchrodingersCattt:
Thanks for the clarification. It sounds like the list form is intentionally kept only for backward compatibility, while `run_mdata_arginfo()` represents the canonical machine schema for external tools. We’ll handle list-form compatibility downstream if needed.

